PR #1, opened 2015-08-22
Option to open in browser by entering a number

jeremija:
Hi there, I added the ability to open a specific result in browser by entering an integer instead of `n` or `p`.

### DIFF
--- a/google
+++ b/google
@@ -52,6 +52,7 @@ class GoogleParser(HTMLParser.HTMLParser):
         self.handle_starttag = self.main_start
         self.handle_data = self.main_data
         self.handle_endtag = self.main_end
+        self.results = []
     def main_start(self, tag, attrs):
         if tag == "li" and len(attrs) > 0 and attrs[0] == ("class", "g"):
             self.title = ""
@@ -85,7 +86,8 @@ class GoogleParser(HTMLParser.HTMLParser):
             if marker >= 0:
                 self.url = self.url[:marker]
 
-            print_entry(self.title, self.url, self.text)
+            index = len(self.results) + 1
+            self.results.append(Result(index, self.title, self.url, self.text))
             self.handle_starttag = self.main_start
             self.handle_data = self.main_data
             self.handle_endtag = self.main_end
@@ -121,34 +123,55 @@ class GoogleParser(HTMLParser.HTMLParser):
             self.handle_endtag = self.li_end
 
 
-# Functions
-def print_entry(title, url, text):
-    # Clean up URL for cases like %22 changed to %2522
-    url = unquote(url)
+class Result:
+    def __init__(self, index, title, url, text):
+        self.index = index
+        self.title = title
+        self.url = url
+        self.text = text
+    def print_entry(self):
+        index = self.index
+        title = self.title
+        url = self.url
+        text = self.text
 
-    # Open the URL in a web browser if option -j was specified.
-    if openUrl:
-        # If empty URL, like first result of 'define hello', just print the result
-        if url == "":
-            print("Showing in console as URL is empty\n")
+        # Clean up URL for cases like %22 changed to %2522
+        url = unquote(url)
+
+        # Open the URL in a web browser if option -j was specified.
+        if openUrl:
+            # If empty URL, like first result of 'define hello', just print the result
+            if url == "":
+                print("Showing in console as URL is empty\n")
+            else:
+                webbrowser.open(url)
+                conn.close()
+                sys.exit(0)
+        # Print the title and the URL.
+        if colorize:
+            print("\x1B[1m\x1B[36m", index, "\x1B[92m", title,
+                  "\x1B[0m\n\x1B[93mURL:", url, "\x1B[39m")
         else:
-            webbrowser.open(url)
-            conn.close()
-            sys.exit(0)
-    # Print the title and the URL.
-    if colorize:
-        print("\x1B[1m\x1B[36m* \x1B[92m", title, "\x1B[0m\n\x1B[93mURL:", url, "\x1B[39m")
-    else:
-        print("* ", title, "\nURL:", url)
-    # Print the text with truncating.
-    col = 0
-    for w in text.split():
-        if (col + len(w) + 1) > columns:
-            col = 0
-            print()
-        print(w, end=' ')
-        col += len(w) + 1
-    print("\n")
+            print(index, title, "\nURL:", url)
+        # Print the text with truncating.
+        col = 0
+        for w in text.split():
+            if (col + len(w) + 1) > columns:
+                col = 0
+                print()
+            print(w, end=' ')
+            col += len(w) + 1
+        print("\n")
+    def open(self):
+        webbrowser.open(self.url)
+
+# Functions
+def is_int(string):
+    try:
+        int(string)
+        return True
+    except:
+        return False
 
 def usage():
     print("Usage: google [OPTIONS] KEYWORDS...")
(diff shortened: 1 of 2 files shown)


jarun:
Awesome change! And you read my mind in the "numbers instead if asterisks" part. Can we make this a little more user-friendly? And a few more suggestions...

1. When the user enters the number to open the link in the browser, but do not exit, fall back to the same prompt. He might want to open more links or keep navigating.

2. The following:
print "  1-10     press a number and Enter to open that result in browser\n"
could be:
print "  1-N     press a number and Enter to open that result in browser\n"

3. Description for point 2 need to be added in the manpage too.

4. I can see some conflicts from this pull request now. Probably due to the last pull request I merged just now.

jeremija:
> 1. When the user enters the number to open the link in the browser, but do not exit, fall back to the same prompt. He might want to open more links or keep navigating.

I agree, I'll take a look.

> 2. The following:
> print " 1-10 press a number and Enter to open that result in browser\n"
> could be:
> print " 1-N press a number and Enter to open that result in browser\n"

If we make the results from 1-N, I believe it would make sense to disable the `p` (previous) command because the previous page will be in the terminal output anyway. And right now it doesn't really go back if `-s` is used.

I say we make the `Enter` without any command load next results, `int` opens a web browser, and `Ctrl-C` exits.

> 3. Description for point 2 need to be added in the manpage too.

Sure.

> 4. I can see some conflicts from this pull request now. Probably due to the last pull request I merged just now.

Will rebase to master.

jarun:
Regarding point 2:

I need to take a look why it is not going back if -s is used. Even if -s is used, I implemented with the first result as base, may be there's a glitch. Thanks for pointing it out! This is also the reason why p cannot be removed, even if the user starts at 50th result, he might want to check out the 40th results by navigating back.

The N comes into picture when -n is used, for example, if '-n 15' is used, the index will range from 1 to 15, so it is not limited to 10.

When the user inputs n or p, please reset the index as if the next set of results is a fresh set. This will reduce complexity of bookeeping and the user doesn't need to scrollback a lot. The idea is to treat each page of results as a new page.

Thank you for the valuable insights :).

jeremija:
@jarun please see the changes.

jarun:
I need some pointers from you on "And right now it doesn't really go back if -s is used." in your second comment. I just checked the following:

$ ./google  -s 20 hello world

Search URL: https://www.google.com/search?<b>start=20</b>&q=hello+world

*  BUMP OF CHICKENHello,world! - YouTube 
URL: http://www.youtube.com/watch?v=rOU4YiuaxAM
...
Press 'n' or 'p' and Enter to continue... p

     *****     *****     *****     *****     

new url:  https://www.google.co.in/search?<b>start=10</b>&q=hello+world&gws_rd=cr&ei=MJjbVfHcJJSBuwTJnpCgAw
*  Hello, world! - The complete C# Tutorial 
URL: http://csharp.net-tutorials.com/basics/hello-world/ 
A chapter on Hello, world! in the complete Microsoft C# .NET tutorial using Visual Studio Express 2012.
...
Press 'n' or 'p' and Enter to continue... p

     *****     *****     *****     *****     

new url:  https://www.google.co.in/search?<b>start=0</b>&q=hello+world&gws_rd=cr&ei=MJjbVfHcJJSBuwTJnpCgAw
*  Hello, World! program - Wikipedia, the free encyclopedia 
URL: https://en.wikipedia.org/wiki/%22Hello,_World!%22_program 
A Hello, World! program is a computer program that outputs Hello, World! (or some variant thereof) on a display device. Because it is typically one of
the...

- where I can see that -s (start) is changing and is reflected in the results... did I miss something here?

jeremija:
Yes, you're right, it does work. I just tested again both with master and this branch. Last time I tried it looked like I was seeing the first page all over again. I must have done something wrong before.

jarun:
You'll see the first page over and over if you'll already hit the first page ;).

jarun:
I could see some issues:
Test 1
-------
Press 'n', 'p', or number and Enter to continue... 6
Press 'n', 'p', or number and Enter to continue... Fontconfig error: Cannot load default config file
9
Press 'n', 'p', or number and Enter to continue... Created new window in existing browser session.

arun@Tux~/D/google-cli> // google-cli exits

Test 2
-------
Press 'n', 'p', or number and Enter to continue... 6
Press 'n', 'p', or number and Enter to continue... Fontconfig error: Cannot load default config file
8
Press 'n', 'p', or number and Enter to continue... 7
Press 'n', 'p', or number and Enter to continue...
// no exit

I tested with Yandex browser beta. In the first test, the utility exited when it opened a new tab. In the second case, it kept working as I closed the browser each time after opening a URL.

Then, continuing with second test, I pressed n and got an error:
Test 3
-------
Press 'n', 'p', or number and Enter to continue... 7
Press 'n', 'p', or number and Enter to continue... n

     *****     *****     *****     *****     

Traceback (most recent call last):
  File "./google", line 310, in <module>
    results = fetch_results()
  File "./google", line 274, in fetch_results
    resp = conn.getresponse()
  File "/usr/lib/python2.7/httplib.py", line 1051, in getresponse
    response.begin()
  File "/usr/lib/python2.7/httplib.py", line 415, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python2.7/httplib.py", line 379, in _read_status
    raise BadStatusLine(line)
httplib.BadStatusLine: ''

jarun:
Does the following issue seem similar?
https://github.com/Schnouki/git-annex-remote-hubic/issues/3

jarun:
Another Chrome related problem:
Press 'n', 'p', or number and Enter to continue... 6
Press 'n', 'p', or number and Enter to continue... [1:2:0825/042917:ERROR:channel.cc(300)] RawChannel read error (connection broken)
[1:2:0825/042917:ERROR:channel.cc(300)] RawChannel read error (connection broken)

jeremija:
> Press 'n', 'p', or number and Enter to continue... Created new window in existing browser session.
>
> arun@Tux~/D/google-cli> // google-cli exits

It seems you pressed an enter here and the program exited because of empty string?

> browser error messages

The problem is that `webbrowser.open()` pipes the output of `stderr` to console. We could use `xdg-open` with `subprocess.Popen` and disable the stderr, but this would then only work on linux. I'm not sure what else could be done here.

> `httplib.BadStatusLine`

I cannot reproduce this. The only thing that crosses my mind is that you experienced a connection failure. But this should have nothing to do with this pull request.

jarun:
I guess I did press enter. I guess you are right about the other cases. I think making a note about using FF should be the right way. I will keep fixing the issues that I see during my tests as well.
Thanks a lot for this valuable change.

Would you mind resolving the conflicts? I believe those came from Python 3 support changes I merged this morning.

jeremija:
> I guess I did press enter. I guess you are right about the other cases. I think making a note about using FF should be the right way. I will keep fixing the issues that I see during my tests as well.
> Thanks a lot for this valuable change.

You're welcome. I'd like to recommend you to use a linter such as [flake8](https://pypi.python.org/pypi/flake8) to verify the syntax. It has sane defaults, configurable rules and makes it a lot of easier for multiple people to use the same coding style.

> Would you mind resolving the conflicts? I believe those came from Python 3 support changes I merged this morning.

I believe that's it. I tested it with both python versions and it seemed to work.

jarun:
Thanks for the flake8 suggestion. Being a C/C++ dev with a customized vim I am indeed having a hard time keeping a tab of the Python syntax. ;)

And thank you so much for your valuable time on this new feature!

jarun:
Merged.

jeremija:
> Thanks for the flake8 suggestion. Being a C/C++ dev with a customized vim I am indeed having a hard time keeping a tab on the Python syntax. ;)

If you're using [syntastic](https://github.com/scrooloose/syntastic), you an easily set your python linter via: `let g:syntastic_python_checkers = ['flake8']` and then you can run it with `:SyntasticCheck()`. I'm sure there are other plugins which do the same thing.

jarun:
I'll check it out!

jarun:
$ /usr/bin/python3 -m pyflakes google
works for me. Thank you for the suggestion!

jeremija:
You're welcome. Glad you got it working! BTW I believe flake8 does more than pyflakes - see [here](https://flake8.readthedocs.org/en/2.0/).

jarun:
True!

PR #3, opened 2015-08-22
Fix of mixed Tabs and spaces + change check on True/False

Narrat:
Two lines got tabs and spaces mixed (line 266 and 290).
And checking on True or False should be done via the keyword "is" and "is not".

The rest is removing of spaces where none should be

Edit: Split it in two commits. If just one of changes is of interest

### DIFF
--- a/google
+++ b/google
@@ -196,7 +196,7 @@ try:
         elif opt[0] == "-j":
             # Option -j
             openUrl = True
-            if num == None:
+            if num is None:
                 num = "1"
         elif opt[0] == "-f":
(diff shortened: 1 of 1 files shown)


jarun:
Thank you for the contributions!

jarun:
Hi Narrat,

I noticed in your other issue that you ported the tool to Python 3 once. Would you mind spending some time to make the current tool Python 3 compatible? I would love to support both 2.7.x and 3.x versions.

Narrat:
Yes, of course. I can do that

PR #4, opened 2015-08-23
Work additionally with Python 3.x

Narrat:
Changes:
 * use the print function from Future. Print statements would have worked with parenthesis around it, but not if the , param was used (print("", end='')).
 * Change in string and input handling made it necessary to differ between py versions
 * Module loading. Some module names changed. Catched via trial and error

__________
Necessary changes to make it work with py2 and py3.
Please test it thoroughly before merging in case I forgot something to change. For me it worked with py2 and py3, but you never know

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # google-cli
-  
+
 ![Screenshot](google-cli.png)
-  
+
 Copyright (C) 2008 Henri Hakkinen
 
 Modified (2015) by Arun Prakash Jana <engineerarun@gmail.com>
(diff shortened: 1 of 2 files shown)


Narrat:
Because of encoding issues that will arise I prepared another commit and branch, but I wait until this got merged, or not.

Commit: https://github.com/Narrat/google-cli/commit/429dccac6ea5e1a390faed47de867f421a6f7c56

jarun:
Seems clean in my tests. Thank you!

jarun:
Please raise the pull request for the encoding change. Just make sure that it doesn't break the existing search options (-s, -n, -f, -t) and navigation.

PR #7, opened 2015-08-25
Force UTF8 request and response

Narrat:
Else there will be wrong decoded chars

_______

Noticed no problems with the options for both python versions.
All working well for me, and issues like that https://github.com/Narrat/translate/issues/7 shouldn't become a problem.

### DIFF
--- a/google
+++ b/google
@@ -222,7 +222,7 @@ except GetoptError as e:
     sys.exit(1)
 
 # Construct the query URL.
-url = "/search?"
+url = "/search?&ie=UTF-8&oe=UTF-8&"
 
 if start is not None:
     url += "start=" + start + "&"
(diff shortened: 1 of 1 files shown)


jarun:
Thank you for this change!

jarun:
Any idea how we can show %22 as " (double quotes)?
For example:
https://en.wikipedia.org/wiki/%22Hello,_World!%22_program
should be
https://en.wikipedia.org/wiki/"Hello,_World!"_program

Narrat:
If it doesn't brake the url (using terminal for opening it, or program) you could try to use unquote from urllib.parse at https://github.com/jarun/google-cli/blob/master/google#L142

jarun:
I am already using it at
https://github.com/jarun/google-cli/blob/master/google#L127

Will give it a rest or now. BTW, I have done some more testing and found no problems with Python 2.7.x or 3.x. Will release version 1.1 soon with your contributions. :)

Narrat:
I obviously missed that too *cough*
Remaining method I currently can think of:
After the unquote put
> url = url.replace("%22", "\"")

jarun:
I will test if it opens correctly in the browser. In the worst case will ave it as it is.

jarun:
Works! Awesome! ;)

PR #8, opened 2015-08-26
indentation fix

shaggytwodope:
Tab was used in place of spaces. This causes an error under python3. Simple fix.


jarun:
Merged. Thanks!

PR #15, opened 2015-10-19
Makefile update for OSX

ibaaj:
```install``` command on OSX hasn't the same behavior as on Linux.
option ```-t``` doesn't exist.


jarun:
Merged. Thank you!

PR #18, opened 2015-12-11
some cleanup based on pep8 stuff

shaggytwodope:
I'd understand if you annoyed with pep8 stuff, but wanted to stop vim from complaining lol.


jarun:
Thanks! I wanted to separate out the func defs for a while myself. :)

PR #24, opened 2016-02-19
Treat EOF at prompt as empty

zmwangx:
Currently, ^D at googler's prompt result in an uncaught `EOFError`, but it doesn't need to be because ^D is a common way to exit shells. This commit makes EOF treated the same as empty input, i.e., exit.


jarun:
Merged. Thank you for your contribution!

PR #25, opened 2016-02-19
README.md: Add Homebrew to the list of installation methods

zmwangx:
Homebrew doesn't have an official online package index, or in other words, https://github.com/Homebrew/homebrew/blob/master/Library/Formula/ is the official index:

```
> brew info googler
googler: stable 2.1 (bottled)
Google Search and News from the command-line
/usr/local/Cellar/googler/2.1 (6 files, 68.6K) *
  Built from source
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/googler.rb
```

I'm linking to http://braumeister.org, the semi-official online package index (semi-official as in independently maintained but linked from official README), because it is more user-friendly to non-Homebrew-developers.

Not sure if you'll be satisfied with this format or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ If you do not want to install, `googler` is standalone:
 <pre>$ chmod +x googler
 $ ./googler ...</pre>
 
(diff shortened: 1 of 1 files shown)


jarun:
Works fine for me. Thank you!

PR #27, opened 2016-02-20
Log full HTML response to a tmpfile when debug is on

zmwangx:
The full HTML response body is the single most valuable reproducer when something goes wrong during the handling of that response.

This commit writes it to a tmpfile (usually under `$TMPDIR`, with the `googler-response-` prefix) and prints the path to console when debug (`-d`) is on:

    [DEBUG] Response body written to '/tmp/googler-response-oqY18e'.

so when asked a user could supply that log to facilitate debugging.

### DIFF
--- a/googler
+++ b/googler
@@ -27,6 +27,7 @@ import webbrowser
 import gzip
 from getopt import getopt, GetoptError
 import readline
+import tempfile
 # Try to load Py3 modules, on error fall back to Py2 module names
 try:
     from io import BytesIO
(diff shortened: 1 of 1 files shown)


jarun:
I am hoping tempfile.mkstemp() plays nice on Windows from the docs! :D

zmwangx:
> I am hoping tempfile.mkstemp() plays nice on Windows from the docs! :D

I believe so. I've certainly written cross-platform tools/libraries before that relies on `tempfile.mkstemp()`, especially in tests.

zmwangx:
Although removing files is always a pain on Windows (can't delete open files), which isn't involved here.

PR #28, opened 2016-02-20
Use logging.debug for printing debug info

zmwangx:
Printing debug info with

```python
if debug:
    print("[DEBUG] blah blah")
```

is both overly verbose and non-idiomatic. It's also not extensible, as in it's hard to change message format, filter message level, change logging destination, etc. Using `logging.debug` is much better.

This commit introduces little user-facing changes, except:

1. Debugging info is now printed to stderr instead of stdout (should be considered an improvement);
2. "`New search URL`" or "`Next URL`" used to not be preceded with "`[DEBUG]`" (not sure why); after this commit they are treated the same as everything else, which improves consistency.

I also took the liberty to add another commit to fix some obvious style issues caught by pylint (I only fixed the hopefully noncontroversial ones), mostly because I don't think they deserve another PR, but they are nevertheless good to see fixed.

### DIFF
--- a/googler
+++ b/googler
@@ -28,6 +28,7 @@ import gzip
 from getopt import getopt, GetoptError
 import readline
 import tempfile
+import logging
 # Try to load Py3 modules, on error fall back to Py2 module names
 try:
     from io import BytesIO
@@ -334,6 +335,15 @@ except GetoptError as e:
     print("googler:", e)
     sys.exit(1)
 
+# Set up logging
+logger = logging.getLogger(__name__)
+shell_formatter = logging.Formatter(fmt='[%(levelname)s] %(message)s')
+shell_handler = logging.StreamHandler()
+shell_handler.setFormatter(shell_formatter)
+logger.addHandler(shell_handler)
+if debug:
+    logger.setLevel(logging.DEBUG)
+
 # Construct the query URL.
 url = "/search?ie=UTF-8&oe=UTF-8&"
 
@@ -351,15 +361,13 @@ if duration is not None:
 baseurl = url
 basestart = start
 
-if debug:
-    print("[DEBUG] Base URL [%s]" % url)
+logger.debug("Base URL [%s]", url)
 
 url += "q=" + quote_plus(keywords[0])
 for kw in keywords[1:]:
     url += "+" + quote_plus(kw)
 
-if debug:
-    print("[DEBUG] Search URL [%s : %s]" % (server, url))
+logger.debug("Search URL [%s : %s]", server, url)
 
 # Get the terminal window size.
 winsz = fcntl.ioctl(sys.stderr, termios.TIOCGWINSZ, "1234")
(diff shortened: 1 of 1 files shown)


googler
```diff
@@ -26,8 +26,8 @@ import struct
 import webbrowser
 import gzip
 from getopt import getopt, GetoptError
-import readline
```
jarun: readline is necessary. Otherwise left/right arrow keys do not work in search prompt.

googler
```diff
@@ -334,6 +334,15 @@ except GetoptError as e:
     print("googler:", e)
     sys.exit(1)
 
+# Set up logging
```
jarun: The full block should be within `if debug`. I don't want to initialize this even if it's not required.

jarun: Is there any way to block the initialization if `-d` option is not used?

googler
```diff
@@ -351,15 +360,13 @@ if duration is not None:
 baseurl = url
 basestart = start
 
-if debug:
-    print("[DEBUG] Base URL [%s]" % url)
+logger.debug("Base URL [%s]", url)
```
jarun: Relating to the earlier comment, I think the `debug` flag is still useful and we can retain it everywhere.

googler
zmwangx: Oops, forgot `readline` is special. Will add back.

googler
zmwangx: Well, kind of... You can wrap the whole logging block in `if debug` if you really want it that way. But then you lose the advantage of `logging`. Is there a particular reason (the other common is rather vague) you want to leave `if debug` everywhere? Because when `debug` is not set, `logger` retains its default level, that is `logging.NOTSET`, so when we subsequently call `logger.debug` the messages are processed by the root logger, the default level of which is `logging.WARNING`, so the debug messages are automatically ignored. On the other hand, if `debug` is set, then `logger`'s level is set to `logging.DEBUG` such that all subsequent `logger.debug` messages go through. So basically we have an implicit `if debug` around each `logger.debug` call already. This is the beauty of `logging`, and still dropping `if debug` everywhere kind of defeats the purpose, then there's little benefit of using `logging` over just `print`. If you don't like the idea of passing to root logger, we can be more explicit and do

```python
if debug:
    logger.setLevel(logger.DEBUG)
else:
    logger.setLevel(logger.WARNING)
```

zmwangx: By the way, another benefit is we can also print errors with `logger.error` and warnings with `logger.warning`, which I didn't get to.

jarun: Frankly, I don't want to execute a single line of code if it's not necessary. The debug mode is for troubleshooting, which I hope is not the regular case with `googler`. As I mentioned earlier, the response body is quite useful but I think at this stage of the project (reasonably matured), and with the few lines of code that it has, introducing a full framework for logging is a overkill. Because it adds to maintenance too. When they changed print from 2.7 to 3.x we needed to adapt to it everywhere. Probably we should avoid that in future by adding features which aren't indispensable.

zmwangx: > introducing a full framework for logging is a overkill.

Yeah probably. However,

> I don't want to execute a single line of code if it's not necessary.

as I said, `if debug` is basically implicit in `logger.debug`.

> Because it adds to maintenance too.

The basic logging interface was stable across Python 2 and 3 (we're not using any advanced feature here), and from all the cues I picked up over the years, there won't be as dramatic a breaking release as Python 3, ever. However, it is a valid argument that using a logging framework adds complexity that is probably not necessary.

Feel free to close this PR.

jarun: > as I said, if debug is basically implicit in logger.debug.
Understood. I meant the initialization too (and without any extra complexity).
> Feel free to close this PR.
Yep, I think we can skip this, at least for now.

Issue #21, opened 2016-02-17
UnicodeEncodeError in print() with KDE Konsole on KaOS


jarun:
The following error is thrown on KaOS when `googler` is invoked from KDE Konsole, (bash shell):

<pre>Our first program will print the classic Traceback (most recent call last):
  File "/usr/local/bin/googler", line 454, in <module>
    results = fetch_results()
  File "/usr/local/bin/googler", line 438, in fetch_results
    r.print_entry()
  File "/usr/local/bin/googler", line 203, in print_entry
    print(w, end=' ')
UnicodeEncodeError: 'ascii' codec can't encode character '\u201c' in position 0: ordinal not in range(128)</pre>
Fish shell works fine though.

jarun:
To fix this issue, add the following in your `~/.bashrc`:
<pre>export PYTHONIOENCODING=UTF-8</pre>

DerpDarako:
Hello, 

I tried to use your fix but it didn't help :

`[romain@b710l0701 ~]googler toto
Traceback (most recent call last):
  File "/usr/local/bin/googler", line 455, in <module>
    results = fetch_results()
  File "/usr/local/bin/googler", line 435, in fetch_results
    parser.feed(gzip.GzipFile(fileobj = StringIO.StringIO(resp.read())).read())
  File "/usr/lib64/python2.7/HTMLParser.py", line 117, in feed
    self.goahead(0)
  File "/usr/lib64/python2.7/HTMLParser.py", line 161, in goahead
    k = self.parse_starttag(i)
  File "/usr/lib64/python2.7/HTMLParser.py", line 308, in parse_starttag
    attrvalue = self.unescape(attrvalue)
  File "/usr/lib64/python2.7/HTMLParser.py", line 475, in unescape
    return re.sub(r"&(#?[xX]?(?:[0-9a-fA-F]+|\w{1,8}));", replaceEntities, s)
  File "/usr/lib64/python2.7/re.py", line 155, in sub
    return _compile(pattern, flags).sub(repl, string, count)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
[romain@b710l0701 ~]echo $PYTHONIOENCODING
UTF-8
`

DerpDarako:
Hello,

I tried to use your fix but it didn't help :

```
[romain@b710l0701 ~]googler toto
Traceback (most recent call last):
  File "/usr/local/bin/googler", line 455, in <module>
[...]

PR #30, opened 2016-02-23
Fix failure when sys.stderr is not a tty (mostly for non-interactive invocation)

zmwangx:
I became aware of this the hard way when I was writing a test case for the [Homebrew formula](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/googler.rb). I didn't report at that time because this related to how the project owner defines the project. Now that I've been honored as a collaborator, I think I might be able to influence the project a little bit more substantially. So here it goes.

---
Currently, `fcntl.ioctl` fails with `IOError` (`OSError` on Python 3.x, where `IOError` is an alias for `OSError`) when sys.stderr is not a terminal. A very easy way to test:

```python-traceback
> googler hello 2>>(tee /dev/null) </dev/null
Traceback (most recent call last):
  File "./googler", line 365, in <module>
    winsz = fcntl.ioctl(sys.stderr, termios.TIOCGWINSZ, "1234")
IOError: [Errno 25] Inappropriate ioctl for device
```

Although googler is mostly terminal oriented, we should not preclude noninteractive use (it is an entirely conceivable use case to query googler with a set of keywords and grep out all URLs in the output); and even for interactive use, `2>/dev/null` leading to an error isn't nice UX either, especially considering that googler prints to stdout for the most part. This failure also precludes any meaningful automated testing.

This commit removes the aforementioned limitation, and also fixes a related comment (hard wrap != truncate, the latter implies loss of information).

---

If this one is merged, then the discussion can be taken to the next level: CI. I believe we'll greatly benefit from using a free CI, say Travis, to do automated testing against multiple Python versions and all TLDs, as well as a set of randomized queries to shake out both coding oversight (say, wrote something that's incompatible with Python 3.2) and actual problems like the one in https://github.com/jarun/googler/issues/21#issuecomment-187255335 with `-c fr`.

The test script can be a simple shell script, and once the CI is set up (which takes like two minutes), that won't be any extra maintenance burden, except we'll be alerted when something breaks, and we'll be able to stop regressions at bay.

### DIFF
--- a/googler
+++ b/googler
@@ -195,7 +195,7 @@ class Result:
                   "\x1B[0m\n\x1B[93m%s\x1B[39m" % url)
         else:
             print("", index, title, "\n%s" % url)
-        # Print the text with truncating.
+        # Hard wrap text if the number of columns is available.
         if columns > 0:
             col = 0
             for w in text.split():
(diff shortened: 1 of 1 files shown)


jarun:
Thank you! Yes, Travis integration is something I had in mind. The automated tests would add great value.

zmwangx:
And the problem with this PR is?

jarun:
Ahh sorry! Done by mistake.

zmwangx:
No problem!

jarun:
Merged! :)

PR #31, opened 2016-02-23
Automated test script and Travis CI integration

zmwangx:
Just to give you something to review. The automated test script is hopefully both reasonably robust and user friendly. Give it a spin:

    ./test

or

    ./test -q

which is quieter. Or read the brief help text first (./test -h).

Some initial builds on my fork can be found here: https://travis-ci.org/zmwangx/googler/builds. Unfortunately I spammed thousands of queries in the first build, and was subsequently blocked. (Although I've updated the script to only generate 100 — this number is customizable — queries.) There seems to be nothing I could do except to wait it out.

PR #32, opened 2016-02-23
[WIP] Automated test script and Travis CI integration

zmwangx:
#31 was prematurely merged. There was still review and rebase to be done. I have since fixed a few issues, and more is on the way. I say we should hard rewind the merge commit (isn't really an issue for users because the merge commit added `test` and `.travis.yml` both of which are not user-oriented), and work on this branch until it is in a ready state. I have tagged this as "do not merge" to prevent accidental merging.

Currently all Python 3 tests pass on Travis, but for some reason Python 2.7 tests all fail (I was running in quiet mode, and all the reruns look just fine, so I'm not really sure what caused the first runs to exit with a nonzero status). It'll take some time for me to investigate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: python
+python:
+  - "2.7"
+  - "3.2"
+  - "3.3"
+  - "3.4"
+  - "3.5"
+  - "nightly"
+matrix:
+  allow_failures:
+    - python: "nightly"
+script:
+  - PYTHONIOENCODING=utf-8 PATH=$PWD:$PATH ./test --ci
--- a/test
+++ b/test
@@ -1,0 +1,149 @@
+#!/usr/bin/env bash
+
+set -e
+
+declare quiet exitcode
+
+quiet=0
+ci=0
+exitcode=0
+while [[ $1 == -* ]]; do
+    case $1 in
+        --ci)
+            ci=1
+            quiet=1
+            ;;
+        -h|--help)
+            cat <<'EOF'
+Usage: test [options]
+
+Run automated tests of googler(1). googler(1) is expected on $PATH.
+
+Requires shuf(1) from coreutils and /usr/share/dict/words.
+
+Options:
+    --ci
+        Same to --quiet, except LF is used instead of CR when printing progress
+        information. The reason is that CI logs are typically line buffered, so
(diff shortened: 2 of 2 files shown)


zmwangx:
All green (for real) now! https://travis-ci.org/zmwangx/googler/builds/111304474

Apart from bug fixes, I have improved the test script to print errors in color, and print exit status when a test fails.

jarun:
I'll wait till you remove the "do not merge" label. Does this indicate that our hack on user agent works fine? Worked for me from India.

zmwangx:
> I'll wait till you remove the "do not merge" label.

Are you okay with hard winding the merge commit of #31?

```sh
> git checkout master
> git reset --hard master~1 # or b595687
> git push -f origin master
```

That way we get rid of the accidentally dirty history, and this patch would apply cleanly. (I've already rebased and this PR is already in a clean and mergeable state bar #31. I will read again to make sure there are no obvious pitfalls.)

> Does this indicate that our hack on user agent works fine?

Bottom: it shouldn't hurt. However, I don't particularly like it because

1. It's unnecessary most of the time;
2. On second thought, I seriously doubt a UA trick — which is usually enough to fool smaller websites — is able to fool Google, Alexa rank *#*1 and holder of world's data. I don't think their spam detection is as naive as "you seem to be a browser, so I'll let you through."

zmwangx:
I pushed some other changes to improve the UI. Most notably, there's now a `--ci` option so that progress info is available in Travis in real time. Example: https://travis-ci.org/zmwangx/googler/jobs/111330695. Basically, I was using CR in quiet mode so that `n/100` is updated on the same line, but Travis (and most other CIs, I suppose) is line-buffered, so I have to use LF instead.

jarun:
Please go ahead with the hard reset. You already have commit permission to master.

zmwangx:
> Please go ahead with the hard reset. You already have commit permission to master.

Done, just wanted to get your approval since rewinding is usually frowned upon.

Now this PR is clean, mergeable, and green. I have removed the "do not merge" label from both here and #33.

jarun:
> On second thought, I seriously doubt a UA trick — which is usually enough to fool smaller websites — is able to fool Google

As I mentioned, I am unable to reproduce it with/without UA. We need active help from the issue reporter.

PR #33, opened 2016-02-23
googler.1: Correct description of option -C

zmwangx:
`-C` is for disabling color output, not the other way round.


zmwangx:
Please defer merging until #32 is processed, which might involve rewriting the history.

jarun:
Thank you!

PR #34, opened 2016-02-25
Refresh README.md

zmwangx:
Please comment on the changes that you find unacceptable, and I'll fix/revert those.

---

Changes I cared to remember:

* Use proper Markdown instead of a rather awkward and inconsistent mix of Markdown and HTML that

    1. Defies tooling (e.g., using Pandoc to generate groff from it);
    2. Doesn't help with readability;

* Completely reworked installation instructions — hopefully more logical now. Other than reorganization, here are the changes in substance:

    1. `sudo` is not always needed (for instance, on my personal OS X system I've changed the mode of `/usr/local` so that it belongs to me and I write access — which is the default of Homebrew, by the way), so leave that bit to the user with a suitable explanation;
    2. Added instructions on `PREFIX`;
    3. Removed the bit about `chmod +x googler`; `googler` already has mode 755 in this repo;
    4. Added a link to Linuxbrew;

* Removed references to `~/.bashrc` because:

    1. Not everyone uses bash;
    2. Even for bash users, `~/.bashrc` could be a non-solution because an interactive login session only reads from `~/.bash_profile` (or fallbacks), not `~/.bashrc`, so referring to `~/.bashrc` only adds confusion;
    3. Most importantly, rather than giving quick instructions to users (oh, just add this line to `~/.bashrc` and you're good), I'd rather see them do independent research on aliases or environment variables if they have never encountered them — I (used to) answer questions on the bash and zsh tags of StackOverflow, and I've seen way too many problems stemming from users mindlessly following half-baked instructions with zero understanding of shell runcoms, and later turning their runcoms into an unsalvageable mess. Fish shell is an outlier that is not remotely POSIX, so specific instructions are fine;

* Removed the note "Google ignores searches for negative numbers" since it's misleading: it's not that Google ignores negative numbers, but rather, Google search allows inline logic, and -*word* means to exclude *word*;

* Removed the warning "`2>&1` would error as the console geometry is calculated from stderr", since we already removed that restriction (softwrapping provided by terminal emulators is used in that case; not as nice as hardwrapping on word boundaries, but infinitely better than erroring out);

* Regarding "Look up n, p, g co or a number at navigation prompt": Removed the whole printout, because it's not really necessary (I believe users could understand what's being said even without that visual context), and who knows if the format of that prompt will ever change and outdate this entry. It's usually bad to be way too specific in docs because docs can hardly keep up with development ;)

* Some minor improvements here and there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,13 +3,28 @@
 ![Screenshot](googler.png)
 
 `googler` is a command-line power tool to search Google (Web & News) from the terminal. It shows the title, URL and text context for each result. Results are fetched in pages. Next or previous page navigation is possible using keyboard shortcuts. Results are indexed and a result URL can be opened in a browser using the index number. Supports sequential searches in a single instance.
-  
-`googler` is GPLv3 licensed. It doesn't have any affiliation to Google in any way.  
-  
+
+`googler` is GPLv3 licensed. It doesn't have any affiliation to Google in any way.
+
 Why not use Google provided APIs? Check point 2 in [Notes](#Notes).
 
-If you find `googler` useful, please consider donating via PayPal.  
-<a href="https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&amp;hosted_button_id=RMLTQ76JSXJ4Q"><img src="https://www.paypal.com/en_US/i/btn/btn_donateCC_LG.gif" alt="Donate Button with Credit Cards" /></a>
+If you find `googler` useful, please consider donating via PayPal.
+[![Donate Button with Credit Cards](https://www.paypal.com/en_US/i/btn/btn_donateCC_LG.gif)](https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&amp;hosted_button_id=RMLTQ76JSXJ4Q)
+
+# Table of contents
+
+- [Features](#features)
+- [Installation](#installation)
+    - [Installing from this repository](#installing-from-this-repository)
+        - [Installing to (or uninstalling from) a prefix](#installing-to-or-uninstalling-from-a-prefix)
+        - [Running as a standalone utility](#running-as-a-standalone-utility)
+    - [Installing with a package manager](#installing-with-a-package-manager)
+- [Usage](#usage)
+    - [Configuration file](#configuration-file)
+- [Examples](#examples)
+- [Troubleshooting](#troubleshooting)
+- [Developers](#developers)
+- [Notes](#notes)
 
 # Features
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,113 +47,163 @@ If you find `googler` useful, please consider donating via PayPal.
...
+    $ git clone https://github.com/jarun/googler/
+    $ cd googler
+
+Otherwise, download the latest [stable release](https://github.com/jarun/googler/releases/latest) or [development version](https://github.com/jarun/googler/archive/master.zip"), unpack the archive, and cd into the directory.
+
+Next, run
+
+    $ make install
+
+which installs to `/usr/local` by default. You might need to prepend `sudo` to the command above if `/usr/local` is only writable by root. To install to a different prefix, run
```
jarun: Let's skip this explanation and keep it with sudo as /usr/local is writable by default.

README.md
```diff
@@ -47,113 +47,163 @@ If you find `googler` useful, please consider donating via PayPal.
...
+
+    $ PREFIX=/path/to/prefix make install
```
jarun: Add a note below this that user might have to use sudo depending on his permissions on the target installation directory.

README.md
```diff
@@ -47,113 +47,163 @@ If you find `googler` useful, please consider donating via PayPal.
...
+
+To remove, run
+
+    $ make uninstall
+
+Again, you might need `sudo` and/or `PREFIX=`.
```
jarun: Please keep the original command with sudo.

README.md
```diff
@@ -47,113 +47,163 @@ If you find `googler` useful, please consider donating via PayPal.
...
+
+### Don't feel like installing to a prefix?
```
jarun: Please keep the sub-section header as:
Run as a standalone utility

README.md
```diff
@@ -47,113 +47,163 @@ If you find `googler` useful, please consider donating via PayPal.
...
+
+`googler` is a standalone executable. From the containing directory:
+
+    $ ./googler
+
```
jarun: Please provide a link to the latest package download manually as a zip package (from 2. below) and add it here. This is for impatient people who don't want to open a package manager and don't have git installed on the system.

README.md
```diff
@@ -47,113 +47,163 @@ If you find `googler` useful, please consider donating via PayPal.
...
+## Installing with a package manager
 
-2. If you do not have git installed:
```
jarun: Please pull this section above for reasons I mentioned in earlier comment.

README.md
```diff
@@ -47,113 +47,163 @@ If you find `googler` useful, please consider donating via PayPal.
...
+
+        $ googler -C hello world < input > output
+
+    Note that `-C` is required to avoid printing control characters (for colored output).
+
+13. **Piping** `googler` output:
+
+        $ googler -C hello world | tee output
+
+14. More help:
```
jarun: Please make "help" bold.

README.md
```diff
@@ -47,113 +47,163 @@ If you find `googler` useful, please consider donating via PayPal.
...
 # Troubleshooting
 
-1. If print() throws the following error complaining about handling Unicode with `ascii` codec:
-<pre>UnicodeEncodeError: 'ascii' codec can't encode character '\u201c' in position 0: ordinal not in range(128)</pre>
-add the following to your `~/.bashrc`:
-<pre>export PYTHONIOENCODING=UTF-8</pre>
-If you use fish shell, add the following in `~/.config/fish/config.fish`:
-<pre>set -x PYTHONIOENCODING UTF-8</pre>
-Ref issue [#21](https://github.com/jarun/googler/issues/21).
+1.  If `print()` throws the following error complaining about handling Unicode with `ascii` codec:
```
jarun: An extra space before If, doesn't show in output though.

README.md
```diff
@@ -161,10 +211,11 @@ Ref issue [#21](https://github.com/jarun/googler/issues/21).
...
 2. Resurrected and maintained (2015 -) by [Arun Prakash Jana](mailto:engineerarun@gmail.com)
 3. [Zhiming Wang](https://github.com/zmwangx)
-  
+
 Special thanks to [jeremija](https://github.com/jeremija), [shaggytwodope](https://github.com/shaggytwodope) and [Narrat](https://github.com/Narrat) for their contributions and efforts in spreading `googler`.
 
-<h1 id="Notes">Notes</h1>
+# Notes
 
 1. Initially I raised a pull request but I could see that the last change was made 7 years earlier. In addition, there is no GitHub activity from the original author [Henri Hakkinen](https://github.com/henux) in past year. I have created this independent repo for the project with the name `googler`. Would love to push the changes back to original repo if the author contacts. I retained the original copyright information.
```
jarun: Let's drop the line: Would love to push the changes back to original repo if the author contacts.
I won't enjoy it much at this stage. Anyone can fork and modify on their own.

README.md
```diff
@@ -161,10 +211,11 @@ Ref issue [#21](https://github.com/jarun/googler/issues/21).
...
+
 2. Google provides a search API which returns the results in JSON format. However, as per my understanding from the [official docs](https://developers.google.com/custom-search/json-api/v1/overview), the API issues the queries against an existing instance of a custom search engine and is limited by 100 search queries per day for free. In addition, I have reservations in paying if they ever change their plan or restrict the API in other ways. So I refrained from coupling with Google plans & policies or exposing my trackable personal custom search API key and identifier for the public. I retained the browser-way of doing it by fetching html, which is a open and free specification.
```
jarun: Thanks for the remarkable improvement of the readme ;)

README.md
zmwangx: Switched to "Running as a standalone utility" to be consistent with other section/subsection names, because we were using nouns or present participles throughout.

README.md
zmwangx: I don't quite understand this because the links are already given above; moreover, the download options should be independent from whether installing to a prefix or running from the repo dir — actually that's the most important reason I restructured the installation section. Anyway I (hopefully) clarified the logical relations between steps even more in the latest version: https://github.com/zmwangx/googler/tree/readme-refresh#installing-from-this-repository.

jarun: Sorry, I missed the pervious link.


Regards,

Arun

http://tuxdiary.com/

README.md
zmwangx: This is intentional, so that this paragraph lines with the following paragraphs under the same list item. Four-space indents are required for some Markdown parsers, a rather unfortunate fact (CommonMark, a Markdown standardization effort, which I'm an advocate for, removes that annoying restriction and only requires you to indent enough to be unambiguous, but the semistandard is not recognized accross the board).

zmwangx: Happy to remove the space if you prefer that way.

zmwangx:
I believe I have either addressed or responded to each concern.

zmwangx:
I just added a table of contents to the README for better navigation. https://github.com/zmwangx/googler/tree/readme-refresh#table-of-contents

jarun:
Thank you very much!

jarun:
Did you come across [buku](https://github.com/jarun/buku) yet? Care to try the exact word matching algorithm (check the Contributions section)?

zmwangx:
I just took a look. Interesting, but probably not what I need since Chrome's bookmarking + completion and some dedicated shell scripts (e.g., I have a script `docs.py` for opening Python 3 doc pages at docs.python.org, and `freebsdman`, `linuxman`, `openbsdman`, `osxman`, `ubuntuman`, etc. for opening online man pages, so on and so forth) are serving me just fine.

I didn't have time to look at the code and see exactly what you need (and I probably won't have much time in the following month), but I assuming you want to mimic Google Search, here is a battery included script with several algorithms (I wrote it quickly so there might be bugs):

```python
#!/usr/bin/env python3

# Several exact word search algorithms.
#
# All searches are case insensitive. Whitespace is always
# insignificant. Query strings are assumed to be well-formed. Needs
# checking/sanitizing in the real world.

import itertools
import re

QUERY = 'google search'
STRICT_QUERY = '"google books" search'

DATA = [
    'pre  Google  Search  post',
    'pre  Google  Videos  Search  post',
    'pre  Google  Books  Search  post',
    'pre  googlesearch  post',
    'pre  searching  Google  post',
    'pre  foo search  google  bar  post',
    'pre  search  google  books  post',
    'pre  search  books  on  Google  post'
]

# Filtering algorithm 1, based on sets.
def filter1(query, data):
    query_words = set(query.lower().split())
    return [s for s in data if query_words.issubset(set(s.lower().split()))]

# Filtering algorithm 2, based on re word boundary and permutations.
def filter2(query, data):
    query_words = query.split()
    # Deduplicate and attach boundaries
    seen = set()
    query_words = [(r'\b' + w + r'\b') for w in query_words if not (w in seen or seen.add(w))]
    # Construct regex
    criterion = re.compile('(' + '|'.join(['.*'.join(p)
                                           for p in itertools.permutations(query_words)]) + ')',
                           re.I)
    return filter(criterion.search, data)

# Strict filtering algorithm, based on re word boundary and permutations.
def strict_filter(query, data):
    # Parse query string into tokens (in regex)
    query_tokens = []
    current_token = ''
    quote = '"'
    for w in query.split():
        if w.startswith(quote):
            if w.endswith(quote):
                query_tokens.append(r'\b' + w[1:-1] + r'\b')
            else:
                current_token = r'\b' + w[1:] + r'\W+'
        elif w.endswith(quote):
            current_token += w[:-1] + r'\b'
            query_tokens.append(current_token)
            current_token = ''
        else:
            if current_token:
                # Part of a quoted token
                current_token += w + r'\W+'
            else:
                query_tokens.append(r'\b' + w + r'\b')
    # Deduplicate
    seen = set()
    query_tokens = [t for t in query_tokens if not (t in seen or seen.add(t))]
    # Construct regex
    criterion = re.compile('(' + '|'.join(['.*'.join(p)
                                           for p in itertools.permutations(query_tokens)]) + ')',
                           re.I)
    return filter(criterion.search, data)

def test_filter(filterfunc, query, desc=None):
    if not desc:
        desc = filterfunc.__name__
    print(desc)
    print('  Query:\n    %s' % query)
    print('  Results:')
    print('    ' + '\n    '.join(filterfunc(query, DATA)) + '\n')

def main():
    test_filter(filter1, QUERY,
                'Filter algorithm #1 (sets and issubset)')
    test_filter(filter2, QUERY,
                'Filter algorithm #2 (re word boundary + permutations)')
    test_filter(strict_filter, STRICT_QUERY,
                'Strict filter algorithm (re word boundary + permutations)')

if __name__ == '__main__':
    main()
```

PR #36, opened 2016-02-25
README.md: Add Travis build status

zmwangx:
Not sure if this is the best idea since there's a small chance that our tests can fail outside our control (i.e. Google blocking suspicious behavior), but it's usually good to have (rather than completely rely on notification emails which might get ignored amongst a sea of emails), and if we drop `[ci skip]` wisely in commit messages (there's no point in CI testing a commit that updates `README.md`, for instance), the chance will be pretty low.


jarun:
Let's skip this when we know too many queries may get us blocked. And I don't think we can do much about it.

PR #38, opened 2016-02-29
Extensibility & DRY: Factor out repeated conn.request and HTTPSConnection

zmwangx:
Sorry about the delay.

---

The same headers/options are repeated over and over. When the headers/options are modified, there's a pretty big chance that one occurrence would be left out (see b595687 for instance), and repeating a long string of options doesn't help readability either.

This commit strives to solve the aforementioned problem by introducing two helper functions `google_get` and `new_connection`.


jarun:
Thank you!

PR #42, opened 2016-03-08
Prompt and compatibility layer improvements

zmwangx:
Kind of a weird mix for a single PR.

* 2to3 compatibility layer: Use `sys.version_info` directly instead of a `try..except` approach. The imports and compatibility layer could be further cleaned up, I'll address that at some later point.

* Prompt: reduce duplication, and most importantly, use reverse video rather than a blanket red bg which might lead to contrast issues in many color schemes.


jarun:
Thank you!

PR #43, opened 2016-03-09
Improve prompt message

zmwangx:
I just took a closer look at the prompt message and I think it could use some improvements.

1. 'n', 'p' and result number are all pertinent to the current search while keywords start a whole new, unrelated search, so 'n', 'p' and result number should be logically grouped together;

2. "Keywords" is rather confusing, because a user who isn't familiar with feature would naturally assume it's something related to the current search. "New keywords" is hopefully more clear.

The screenshot will have to be updated again if this is accepted...

### DIFF
--- a/googler
+++ b/googler
@@ -374,7 +374,7 @@ def quit(conn):
 def show_omniprompt():
     global colorize
 
-    message = "Enter 'n', 'p', keywords or result number to continue"
+    message = "Enter n, p, result number or new keywords"
     if colorize:
         return raw_input("\x1b[7m%s\x1b[0m " % message)
     else:


zmwangx:
Wait what happened to this branch... I'll take a look.

zmwangx:
@jarun You seem to have rebased and force pushed to master? You forgot to amend the commit message in https://github.com/jarun/googler/commit/9e19f7db3491b28315a4b1617b0026fb01ee0d55 when rebasing. Also, the image is still in the repo, and there's no way to remove it without pushing existing tags off master, so don't think rebasing is worthwhile.

jarun:
I have removed the image and its full history completely. Just rebase back to master.

zmwangx:
I've already rebased this branch on master, but I'm not sure about "removed the image and its full history completely". Will check in a second.

zmwangx:
Okay, the image is gone, but that leaves some dangling empty commits.

jarun:
Regarding the change, how about:
    "Enter n, p, result number or new keywords"

jarun:
I used `git filter-branch` method.

zmwangx:
@jarun `google-cli.png` is still in the repo, you just removed the renamed copy, so there's no real benefit. Also, in rebasing you also retagged v2.1 from d70a479750ce4d555225fb4ffc6471d8fc04dc22 to 9a2bc44049c2bedb9e9b9b6c7de1cb99561978a2, and the checksum of the v2.1 tarball also changed. This is a really destructive move and doesn't work well with package managers at all (for instance, I can't build googler from source with Homebrew now due to SHA-256 mismatch). Could you please push back the original tree? As I said there's no real benefit.

zmwangx:
> "Enter n, p, result number or new keywords"

Sure, I have no strong feeling about the quotes.

jarun:
Then wait, I'll remove google-cli.png too.
The original tree is gone. I followed [this](https://help.github.com/articles/remove-sensitive-data/). Will be more careful in the future.

You might have to update the Homebrew entry.

zmwangx:
> The original tree is gone.

It's not gone, I have it here. I'll push the original master back with your permission. You might also have it if you check your git-reflog, although if you gc'ed they might be gone.

jarun:
OK then. Push it back.

jarun:
Wanted to try this out. But this has been foolish ;). Anyway will try this on another personal pet project :P.

zmwangx:
No problem, should be back on track now. By the way, I seriously recommend [bfg-repo-cleaner](https://rtyley.github.io/bfg-repo-cleaner/) over `git-filter-branch`, although I'd say retagging is bad 99% of the time.

jarun:
Yep, understood the issue when it's no more a personal project. :dancer:

zmwangx:
Rebased this branch again, and removed quotes around 'n' and 'p'.

zmwangx:
Wait, you also meant to remove "to continue", is that correct? (I have no strong feeling either way.)

jarun:
remove the `to continue` at the end too! We already said `Enter` ;).

jarun:
> Rebased this branch again

Awesome! We are rocking!

zmwangx:
Done.

jarun:
Thanks!

PR #44, opened 2016-03-10
Clean up import system and compatibility layer

zmwangx:
See commit message of 51d6a15.

The only functional change is we now use `io.BytesIO` across Python 2 and 3 when gunzipping.

\* Note that now imports are ordered alphabetically so as to make them more easily manageable.

### DIFF
--- a/googler
+++ b/googler
@@ -18,34 +18,42 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import print_function
-import sys
-import os
-import termios
 import fcntl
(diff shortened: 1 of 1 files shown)


jarun:
I'll hold this for next release. We'll release version 2.2 this weekend. There are 71 commits already.
Otherwise it'll be too heavy for me to test and confirm everything works.

Hope that works for you!

zmwangx:
Sure, no hurry.

jarun:
Thank you!

PR #45, opened 2016-03-11
Switch to argparse for argument processing

zmwangx:
<del>I based this branch on #44 because I think that one is basically a no-brainer (of course, let's leave it after cutting a release).</del>

The advantages of `argparse` were discussed [in the Gitter channel](https://gitter.im/jarun/googler?at=56e1f39e9f24605773d713520).

PR #46, opened 2016-03-12
Add support for the complete list of TLDs

zmwangx:
The full list of TLDs and their corresponding domains are scraped from https://en.wikipedia.org/wiki/List_of_Google_domains with the script https://gist.github.com/zmwangx/b976e83c14552fe18b71.

Also did some drive-by updates and improvements to the docs (README, googler.1) in this branch. I based this branch on #45 so that I don't need to redo doc changes once #45 is merged. If #45 for whatever reason is rejected, I'll port the changes to master.

See the respective commit messages for more details, which should be very informative.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,31 +104,37 @@ If you installed to a different prefix, you'll need to specify the same `PREFIX`
 
 # Usage
 
-    $ googler
-    Usage: googler [OPTIONS] KEYWORDS...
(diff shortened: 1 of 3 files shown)


zmwangx:
Rebased on master.

jarun:
Thanks!

PR #48, opened 2016-03-12
Set encoding of standard streams to UTF-8

zmwangx:
Only necessary for Python 2. This way users don't need to have `PYTHONIOENCODING` set to `utf-8`. 

Also removed `PYTHONIOENCODING` from `.travis.yml`.

<del>I based this branch on #44 because it involves the compat layer (the actual change is acfc53c3c9f73775538932aa66427b6ddf99d883, which is very straightforward). Hence blocked on #44.</del>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,4 @@ matrix:
   allow_failures:
     - python: "nightly"
 script:
-  - PYTHONIOENCODING=utf-8 PATH=$PWD:$PATH ./test --ci
+  - PATH=$PWD:$PATH ./test --ci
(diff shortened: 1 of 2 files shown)


zmwangx:
Rebased on master.

jarun:
Thank you!

zmwangx:
@jarun Do you think we should also remove the [troubleshooting entry](https://github.com/jarun/googler#troubleshooting) about ASCII code from README? Or at least for now add a notice that the problem has been resolved in c2df4a0 and will be incorporated into v2.3? Then at a later stage we could phase it out.

jarun:
Yes, I left it earlier because we have just fixed this in dev. Let's shorten it before the next release, by adding the released version (2.3 probably) with the fix.

zmwangx:
Sounds good.

jarun:
Come to gitter, if possible...

PR #49, opened 2016-03-13
Add a few messaging wrappers and switch to these wrappers

zmwangx:
Error/warning/debug wrapper (basically, lightweight alternative to PSL `logging`), plus a shortcut to print to stderr.

Highlights:

1. Now messages and info are strictly on stderr, so stdout is purely for results;

2. Colorize error and warning messages (error red, warning yellow) when colorizing.

### DIFF
--- a/googler
+++ b/googler
@@ -433,6 +433,53 @@ def show_omniprompt():
         return raw_input("%s: " % message)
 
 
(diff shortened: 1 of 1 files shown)


jarun:
Thanks!

PR #50, opened 2016-03-15
More graceful exit on SIGINT

zmwangx:
Currently, sending a SIGINT results in a traceback from an unhandled `KeyboardInterrupt`. However, SIGINT is so common and harmless that we shouldn't make a fuss about it.

This commit installs a graceful SIGINT handler that prints "Interrupted." and exits the program on SIGINT.

### DIFF
--- a/googler
+++ b/googler
@@ -24,6 +24,7 @@ import gzip
 import io
 import os
 import readline
+import signal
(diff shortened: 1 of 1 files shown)


jarun:
;)

PR #51, opened 2016-03-18
googler.1: Add examples from README.md

zmwangx:
I don't know about you, but I'm always grateful to man pages that include usage examples in addition to a million lines of dry option descriptions. And we have examples already, so why not.

I certainly don't enjoy writing groff, by the way. `mdoc` may be a slightly better alternative, but I'll just stick to groff here.


jarun:
Good idea! Thanks!

PR #55, opened 2016-04-10
just a small change

akash-vartak:
Users ought to understand what is it asking for, I couldn't at first, so I made the change.
Also it's a good utility for people like me who use the terminal constantly

### DIFF
--- a/googler
+++ b/googler
@@ -441,7 +441,7 @@ def quit(conn):
 def show_omniprompt():
     global colorize
 
-    message = "Enter n, p, result number or new keywords"
+    message = "Enter: n for next set of results / p for previous set of results / a result number / new keywords for searching"
     if colorize:
         return raw_input("\x1b[7m%s\x1b[0m " % message)
     else:


googler
```diff
@@ -441,7 +441,7 @@ def quit(conn):
 def show_omniprompt():
     global colorize
 
-    message = "Enter n, p, result number or new keywords"
+    message = "Enter: n for next set of results / p for previous set of results / a result number / new keywords for searching"
```
jarun: We maintained a small prompt intentionally. This new string is too large.

akash-vartak:
The short prompt can be confusing to a few.

jarun:
I understand. I am working on a way to handle it... like the `fdisk` prompt.

zmwangx:
> The short prompt can be confusing to a few.

That's why we have help text and man page. You are expected to read the documentation when anything is unclear, and as far as the prompt goes, you only need to read the doc before you won't need any help, ever (well, maybe you'll need a refresher from time to time if you use googler only sparingly). Your approach of expanding it to 111 characters would be very annoying beyond the first time. I personally hate software that tries to make it slightly easier for new users at the cost of usability for regular users.

zmwangx:
@jarun While `?` is nice, I don't think we should take away the existing message. `?` could just be another option: `n, p, result number, new keywords, or ? for help`.

(Look at `git-commit --patch`.)

jarun:
Implemented a more helpful and simpler omniprompt at commit https://github.com/jarun/googler/commit/31cc35a19f265c75c2a7da607520860d8736bd49.

Hope this works well for you!

Closing the pull request. Thanks for bringing this up! :+1:

jarun:
> Enter n, p, result number or new keywords (? for help)

Looks OK?

zmwangx:
> Enter n, p, result number or new keywords (? for help)

Yeah, better than mine.

akash-vartak:
Yeah that's good....

jarun:
Updated at commit ff3ee520a436ca06c05ad10e770c3fde2258fb85.

PR #58, opened 2016-04-11
Move test to subdirectory tests/

zmwangx:
Now that we have a directory structure (`auto-completion/`), why not be a little bit more standard and hide the details of testing harness in `tests/`, even though it's just a single script.


jarun:
Merged.

PR #62, opened 2016-04-22
README.md: Add AUR and Homebrew badges to the top

zmwangx:
This way AUR or Homebrew users won't need to scrolldown to find out whether it's available. (Actually, a reasonably experienced user will just search or even guess from the command line; anyway, it doesn't hurt to make it obvious.)


jarun:
Thanks!

PR #70, opened 2016-05-01
Augment compat layer to improve portability

zmwangx:
See commit message for details.

After this patch, googler can run on Windows given that you change terminal code page to UTF-8 (I figured that one out: `chcp 65001`). You can even get color if you use a better third-party terminal emulator, e.g., ConEmu. Here's a screenshot of googler running in ConEmu:

googler-in-conemu

### DIFF
--- a/googler
+++ b/googler
@@ -19,16 +19,12 @@
 
 from __future__ import print_function
 import argparse
-import fcntl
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #74, opened 2016-05-03
Add a non-interactive mode (--noprompt)

### LINKED ISSUE #72: Suppress Omniprompt
I am automating a bunch of searches and redirecting the results to text files.  I would like the option to suppress the Omniprompt text from the results.

By this I mean I do not want the text "Enter n, p, result number or new keywords (? for help):" to appear at the end of the file.

Also, it seems to wait for user input even when output is redirected.  I'd like the option to suppress the omniprompt entirely.

Thank you.

zmwangx:
Perform the given search and exit. No more prompts and/or interactions.

Good for scripting, but also useful in an interactive shell for those we don't typically look at results on the second page.

Fixes #72.

I also renamed `openUrl` to `open_url` for consistency (noticed this when I was looking at how a I'm feeling lucky search breaks the loop).

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,8 @@ Please substitute `$version` with the appropriate package version.
       -t dN, --time dN      time limit search [h5 (5 hrs), d5 (5 days), w5 (5
                             weeks), m5 (5 months), y5 (5 years)]
       -w SITE, --site SITE  search a site using Google
+      --np, --noprompt      perform search and exit, do not prompt for further
+                            interactions
       -d, --debug           enable debugging
 
     omniprompt keys:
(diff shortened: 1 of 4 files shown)


zmwangx:
@jarun I'm okay with `--noprompt`, but I'm inclined to leave the underlying global variable as `noninteractive`, because in a non-interactive mode we should do more them just not prompt (for instance, we should leave out non-essential info like "... ads skipped").

As for short option, both `-n` and `-N` are taken, and non-interactive is usually associated to `-f` but doesn't apply in this case since our prompt is not a YN prompt and there's nothing "forced". If you have a good letter I'm all ears, but what about `--noprompt` and `--np`?

jarun:
`--noprompt` and `--np` are nice :+1:

zmwangx:
Reworked option names.

jarun:
Thanks!

PR #75, opened 2016-05-03
Add support for JSON output (--json)

zmwangx:
<s>Blocked on #74 (this branch is based on #74).</s>

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ Please substitute `$version` with the appropriate package version.
       -t dN, --time dN      time limit search [h5 (5 hrs), d5 (5 days), w5 (5
                             weeks), m5 (5 months), y5 (5 years)]
       -w SITE, --site SITE  search a site using Google
+      --json                output in JSON format; implies --noprompt
       --np, --noprompt      perform search and exit, do not prompt for further
                             interactions
       -d, --debug           enable debugging
--- a/googler
+++ b/googler
@@ -21,6 +21,7 @@ from __future__ import print_function
 import argparse
 import gzip
 import io
+import json
 import os
 import signal
 import sys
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -157,6 +157,8 @@ Please substitute `$version` with the appropriate package version.
       -t dN, --time dN      time limit search [h5 (5 hrs), d5 (5 days), w5 (5
                             weeks), m5 (5 months), y5 (5 years)]
       -w SITE, --site SITE  search a site using Google
+      --json                output in JSON format; implies --noprompt
+      --noninteractive      perform search and exit, do not prompt for further
```
jarun: Typo perhaps? As per our earlier discussion we are not going to have only `--noprompt`.

googler
```diff
@@ -646,6 +658,9 @@ debug = args.debug
 keywords = args.keywords
 insite = args.insite
 noninteractive = args.noninteractive
+if args.json:
+    output_format = 'json'
+    noninteractive = True
```
jarun: should we couple json and noninteractive? Users can use both the options together if they ust want to dump the results of a single fetch.

zmwangx: JSON output is most likely piped, I don't think anyone would rather read JSON in interactive use. Also, our prompt uses `raw_input` (python2) or `input` (python3) which writes stdout, totally screwing up piping.

There's always `--count` and `--start` to handle pagination.

README.md
zmwangx: I wrote this feature before changing the option name. This is an oversight during rebasing.

zmwangx: Fixed.

googler
jarun: :+1:

zmwangx:
Ping. Any other comments?

jarun:
Nope. Sorry I almost forgot about this PR.

jarun:
Merged. Thanks!

zmwangx:
No problem. I'll add it to the completions.

PR #76, opened 2016-05-11
Rewrite parser logic, and more

zmwangx:
Currently the parser logic is a mess. If you (try to) follow it, you'll realize that it's essentially spaghetti code with convoluted GOTOs. This is nightmare for readability and extensibility (I want to extend the parser to support sitelinks, for instance). Therefore, I've rewritten the parser logic entirely.

The new parser logic is thoroughly documented, complete with rationale. See code comments for details.

Breaking changes:

1. Number of skipped results is no longer shown. It has never been accurate by any means (just a feel-good number, I guess) because few skipped results survive long enough to see the skipping counter incremented; with new, stricter parser logic, even fewer, if any, should survive.

2. Google News -- abstract's back! For JSON output, abstract is now the abstract, and publishing info (publisher and time) are now in a new "metadata" field.

3. Minor: `&sa...` is no longer wrongly stripped from URLs (refer to Google Books results).

4. Minor: Filetype indicators (e.g. [PDF]) are now followed by a space.

### DIFF
--- a/googler
+++ b/googler
@@ -123,7 +123,6 @@ colorize = True             # If True, colorizes the output (option -C)
 duration = None             # Time limit search (option -t) [e.g. h5, d5, w5, m5, y5]
 conn = None                 # Use a single global connection during navigation
 nav = 'n'                   # For user navigation
-skipped = 0                 # Count for skipped ads or blank links
 debug = False               # Print debug logs
 news = False                # Read news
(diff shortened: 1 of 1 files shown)


jarun:
Are we still showing the time in news?

jarun:
Thank you!

PR #78, opened 2016-05-12
Add omniprompt command "o"

zmwangx:
For opening the current search (Google Search or News) in browser. For instance, using `o` within

    googler -d --time w2 --count 5 --lang de google

opens https://www.google.com/search?ie=UTF-8&oe=UTF-8&start=0&num=5&hl=de&tbs=qdr:w2&q=google.

This is very useful for development, and I can also see real world use cases for it.


jarun:
Thanks!

PR #80, opened 2016-05-12
Implement ignore list feature

### LINKED ISSUE #79: site search broken?
```
$ googler -d -c in -n 4 -w imdb.com mr robot
[DEBUG] Version 2.4
[DEBUG] Base URL [/search?ie=UTF-8&oe=UTF-8&start=0&num=4&]
[DEBUG] Search URL [www.google.co.in : /search?ie=UTF-8&oe=UTF-8&start=0&num=4&q=mr+robot+site:imdb.com]
[DEBUG] Response body written to '/tmp/googler-response-b40uq9sd'.

 1 Mr. Robot (TV Series 2015– ) - IMDb
http://www.imdb.com/title/tt4158110/
... who suffers from social anxiety disorder and forms connections through hacking. He's recruited by a mysterious anarchist, who calls himself Mr. Robot.

 2 Elliot played by Rami Malek | Characters & Crew | Mr
http://www.usanetwork.com/mrrobot/cast/elliot

 3 USA Network's Mr. Robot is Filming Again in C
http://amusingthezillion.com/2015/04/19/usa-networks-mr-robot-is-filming-again-in-coney-island/

 4 Mr. Robot (TV Series 2015– ) - IMDb
http://m.imdb.com/title/tt4158110/

 5 Mr. Robot (TV series) - Wikipedia, the free encyclopedia
https://en.m.wikipedia.org/wiki/Mr._Robot_(TV_series)

 6 Mr. Robot (TV Series 2015– ) - Episodes - IMDb
http://www.imdb.com/title/tt4158110/episodes?season=2
GET DISCOVERED. Enhance your IMDb Page. Go to IMDbPro ». |; Help · Login · Register · Login · Mr. Robot (TV Series 2015– ) Poster · Mr. Robot (2015– ) ...

 7 Mr. Robot (TV Series 2015– ) - Episodes - IMDb
http://www.imdb.com/title/tt4158110/episodes
Elliot attempts to hack Vera out of jail in order to save someone he cares about; Tyrell's "game" gets crazy; and Angela digs deeper into her mother's death.

 8 Mr. Robot (TV Series 2015– ) - Full Cast & Crew - IMDb
http://www.imdb.com/title/tt4158110/fullcredits/
Mr. Robot (TV Series 2015– ) cast and crew credits, including actors, actresses, directors, writers and more.

Enter n, p, o, result index or new keywords (? for help) o
[DEBUG] Next URL [/search?ie=UTF-8&oe=UTF-8&start=0&num=4&q=mr+robot+site:imdb.com]

Enter n, p, o, result index or new keywords (? for help)
```

Looks like card results (People who ask) are also showing up.

screenshot

Are we showing these deliberately? I would expect the card results to be shown as indented or not to show at all. Even if we don't show in case of site-specific search, I guess it's OK.

zmwangx:
Fixes #79.

See updated documentation for how it works.

This does add some overhead because the ignore list is matched against each tag, but for a typical page, I measure a mere ~5ms increase in CPU time.

Currently ignored: `div.related-question-pair`.

For the record, the reason I go with a dict for each selector instead of using a real CSS selector is because we simply can't implement CSS selector parsing in a reasonable number of lines.

### DIFF
--- a/googler
+++ b/googler
@@ -131,14 +131,52 @@ _VERSION_ = 2.4             # Current version
 def annotate_tag(annotated_starttag_handler):
     # See parser logic within the GoogleParser class for documentation.
     #
+    # In particular, search for "Ignore List" to view detailed
+    # documentation of the ignore list.
+    #
     # annotated_starttag_handler(self, tag: str, attrsdict: dict) -> annotation
     # Returns: HTMLParser.handle_starttag(self, tag: str, attrs: list) -> None
 
     def handler(self, tag, attrs):
-        annotation = annotated_starttag_handler(self, tag, dict(attrs))
-        if tag not in self.tag_annotations:
-            self.tag_annotations[tag] = []
-        self.tag_annotations[tag].append(annotation)
+        # Get context; assumes that the handler is called SCOPE_start
+        context = annotated_starttag_handler.__name__[:-6]
+
+        # If context is 'ignore', ignore all tests
+        if context == 'ignore':
+            self.insert_annotation(tag, None)
+            return
+
+        attrs = dict(attrs)
+
+        # Compare against ignore list
+        ignored = False
+        for selector in self.IGNORE_LIST:
+            for attr in selector:
+                if attr == 'tag':
+                    if tag != selector['tag']:
+                        break
+                elif attr == 'class':
+                    tag_classes = set(self.classes(attrs))
+                    selector_classes = set(self.classes(selector))
+                    if not selector_classes.issubset(tag_classes):
+                        break
+                else:
+                    if attrs[attr] != selector[attr]:
+                        break
+            else:
+                # Passed all criterions of the selector
+                ignored = True
+                break
+
+        # If tag matches ignore list, annotate and hand over to ignore_*
+        if ignored:
+            self.insert_annotation(tag, context + '_ignored')
+            self.set_handlers_to('ignore')
+            return
+
+        # Standard
+        annotation = annotated_starttag_handler(self, tag, attrs)
+        self.insert_annotation(tag, annotation)
 
     return handler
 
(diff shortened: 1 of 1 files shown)


googler
```diff
@@ -339,6 +403,17 @@ class GoogleParser(HTMLParser.HTMLParser):
 
         self.set_handlers_to('main')
 
+    ### Ignore list ###
+    IGNORE_LIST = [
+        # "People also ask"
+        # Sample query: VPN
+        # Screenshot: https://i.imgur.com/yfcsoQz.png
+        {
+            'tag': 'div',
```
jarun: As I understand from your notes and a quick look at the codes, you are matching each tag against IGNORE_LIST.
For this case, please ignore the `related-question-pair` class while processing inner/outer div. As I see it, we are never going to hit a situation where `related-question-pair` comes outside a div. And it doesn't show up in every result either. So the extra processing in matching against every tag is redundant in most cases.

zmwangx: No, `div.related-question-part` is outside of the topmost result container `div.g`, so you have to watch for that in the `main` scope, which is the scope we're in for the majority of the time. Sure, I can add a conditional to check for it only in the `main` scope, but that means more complexity in the ignore list entry, and honestly, the ignore list is only useful for things that are in the main scope — we can avoid dropping into other scopes in the first place.

Also, as I mentioned, field test shows a consistent increase of <5ms in processing time, that means <0.1s even if we have 20 entries in the ignore list.

jarun: Okies. Thanks for explaining. I was too lazy to open the HTML output. :)

jarun:
Thank you! :)

googler
zmwangx: No problem!

zmwangx: > the ignore list is only useful for things that are in the main scope — we can avoid dropping into other scopes in the first place.

On second thought, this may not be true, but I think it would be better to defer adding the scope test until we really need it (we can certainly recognize a `handler_scope` key in selectors), because ignore list entry tests are pretty lightweight, and an extra scope test for `main` that passes more often than not might not even be worth the saving.

PR #81, opened 2016-05-13
Do not fetch previous page when already at the first page

zmwangx:
If `start` is already 0, attempting to fetch "the previous page" just makes the same request all over again, which is a waste of resources. We just print an error message saying

    Already at the first page.

in this case.

I also made `start` and `num` ints. As always, see commit message for detailed rationale.


jarun:
Thank you!

PR #85, opened 2016-05-22
REPL fixes and improvements

zmwangx:
See individual commit messages, which documented in detail the rationale of changes and behavioral changes.


zmwangx:
An additional note on why that `replace` is unnecessary (and wrong): we already use `url_quote_plus` which quotes ` ` to `+` instead of `%20`.

jarun:
Thank you! :+1:

PR #88, opened 2016-05-23
Makefile improvements

zmwangx:
See individual commit messages for details.

Most notably:

1. Support `PREFIX=/path/to/prefix make install` in addition to `make install PREFIX=/path/to/prefix`;
2. Add targets `install.comp` and `uninstall.comp` for completions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,11 @@ PREFIX ?= /usr/local
 BINDIR = $(DESTDIR)$(PREFIX)/bin
 MANDIR = $(DESTDIR)$(PREFIX)/share/man/man1
 DOCDIR = $(DESTDIR)$(PREFIX)/share/doc/googler
+BASHCOMPDIR = $(DESTDIR)$(PREFIX)/etc/bash_completion.d
+FISHCOMPDIR = $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d
+ZSHCOMPDIR = $(DESTDIR)$(PREFIX)/share/zsh/site-functions
 
-.PHONY: all install uninstall
+.PHONY: all install install.comp uninstall uninstall.comp
 
 all:
 
@@ -17,7 +20,16 @@ install:
 	install -m644 README.md $(DOCDIR)
 	rm -f googler.1.gz
 
+install.comp:
+	install -m755 -d $(BASHCOMPDIR) $(FISHCOMPDIR) $(ZSHCOMPDIR)
+	install -m644 auto-completion/bash/googler-completion.bash $(BASHCOMPDIR)
+	install -m644 auto-completion/fish/googler.fish $(FISHCOMPDIR)
+	install -m644 auto-completion/zsh/_googler $(ZSHCOMPDIR)
+
 uninstall:
 	rm -f $(BINDIR)/googler
 	rm -f $(MANDIR)/googler.1.gz
 	rm -rf $(DOCDIR)
+
+uninstall.comp:
+	rm -f $(BASHCOMPDIR)/googler-completion.bash $(FISHCOMPDIR)/googler.fish $(ZSHCOMPDIR)/_googler
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,7 @@
-PREFIX=/usr/local
-BINDIR=$(DESTDIR)$(PREFIX)/bin
-MANDIR=$(DESTDIR)$(PREFIX)/share/man/man1
-DOCDIR=$(DESTDIR)$(PREFIX)/share/doc/googler
-
+PREFIX ?= /usr/local
+BINDIR = $(DESTDIR)$(PREFIX)/bin
+MANDIR = $(DESTDIR)$(PREFIX)/share/man/man1
+DOCDIR = $(DESTDIR)$(PREFIX)/share/doc/googler
 
 .PHONY: all install uninstall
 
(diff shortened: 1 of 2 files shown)


jarun:
Thank you!

Makefile
```diff
@@ -1,11 +1,12 @@
...
-MANDIR=$(DESTDIR)$(PREFIX)/share/man/man1
-DOCDIR=$(DESTDIR)$(PREFIX)/share/doc/googler
-UNAME_S:=$(shell uname -s)
+PREFIX ?= /usr/local
+BINDIR = $(DESTDIR)$(PREFIX)/bin
+MANDIR = $(DESTDIR)$(PREFIX)/share/man/man1
+DOCDIR = $(DESTDIR)$(PREFIX)/share/doc/googler
+BASHCOMPDIR = $(DESTDIR)$(PREFIX)/etc/bash_completion.d
+FISHCOMPDIR = $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d
+ZSHCOMPDIR = $(DESTDIR)$(PREFIX)/share/zsh/site-functions
```
jarun: We are effective installing the scripts to:
```
/usr/local/etc/bash_completion.d
/usr/local/share/fish/vendor_completions.d
/usr/local/share/zsh/site-functions
```

instead of the intended

```
/etc/bash_completion.d
/usr/share/fish/vendor_completions.d
/usr/share/zsh/site-functions

zmwangx: That's the correct behavior (or at least, one correct behavior). See [GNU Coding Standards on `sysconfdir`](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) or [FHS](https://www.linuxbase.org/betaspecs/fhs/fhs/ch04s09.html).

zmwangx: As for why not `/usr` and `/etc`, you can't simply assume you that have access to them, especially when you accept `PREFIX`. Also, shells are not necessarily installed in `/bin` or `/usr/bin` (my latest Bash, Fish and Zsh are all in `/usr/local`, for instance). I said "Please refer to your shell's manual for instructions on setting up your shell to use completions", so users should figure out paths themselves.

jarun:
> That's the correct behavior (or at least, one correct behavior)

It's useless if it doesn't work. I installed and found that none of the completion scripts work without tinkering.

It's great to have a lot of standard docs and all, but if things don't work out of the box for a user migrating from Windows to Linux, we should abandon those standards.

If you look at the completion paths we are using currently, there's no way to install them in a sensible path with a single PREFIX.

zmwangx:
> I installed and found that none of the completion scripts work without tinkering.

You just assumed that shells are in `/usr`. For me, latest versions of shells are in `/usr/local`, so installing to `/usr` also won't work. Not to mention Bash doesn't have a default completion script directory, you always have to tinker somehow. Maybe the distro's default `bashrc` does that for you, but not a safe assumption to make. I made my points a bit more clear in #95, and added some additional scenarios.

> things don't work out of the box for a user migrating from Windows to Linux, we should abandon those standards.

I fail to see how "migrating from Windows to Linux" is remotely related to this discussion. `install.comp` is not in `all` and is not a standard target, so it's not obvious to begin with, and if you just want to target the low end of the spectrum, people won't even realize this is a thing. Also, I believe being logical beats makeshift hacks that would cause grievance once you dig in. Maybe it's just my perspective — I'm definitely a "power to the power user" guy, and would also like to see more regular folks actually spend a few hours understanding their shells.

jarun:
> For me, latest versions of shells are in `/usr/local`

Custom installed or by default? I am on Ubuntu at the moment. Referred to the default install location.

> migrating from Windows to Linux

I wanted to mean newbies not well-accustomed to Linux yet

zmwangx:
> Custom installed or by default?

Homebrew on OS X defaults to `/usr/local`. OS X ships a horribly outdated bash 4.3.42 (thanks GPLv3) and a pretty recent zsh. It doesn't ship fish. I actually quite like separating system packages from user-installed packages.

Of course on Linux most if not all package managers will install to `/usr`, but for folks like me who like the latest and greatest and aren't afraid of breaking and/or tinkering with stuff (which is why Arch seems to suit me the best among Linux distros), often times you still need to deal with `/usr/local` I suppose, especially if you're on Ubuntu/Debian which is almost always outdated.

> I wanted to mean newbies not well-accustomed to Linux yet

Not having shell completions doesn't really hurt. Most folks, even most advanced users are pretty satisfied with Bash's crappy completions.

zmwangx:
I meant bash 3.2.57 (copied the wrong version).

zmwangx:
Oh, btw I don't think newbies are accustomed to installing from source. When I first started learning the Unix command line years ago, I spent half a year or so building and installing all kinds of stuff from source, overwrote all kinds of system packages and created all kinds of crazy directories in system locations. By the time I figured out the correct ways to do things, I already had the shell figured out. Therefore, I would say people who are competent enough to install from source probably know enough about shells or don't care at all, while other beginners will most likely screw up anyway.

jarun:
> Homebrew on OS X defaults to /usr/local

We're doomed unless we detect the OS or show examples on different OS-es. I'ld say let's skip the auto-completion script installation.

> Oh, btw I don't think newbies are accustomed to installing from source.

With the few lines of steps we provide, if a newbie wants to, he should not face any issues. Till the regular install, uninstall we were good. shell completion installation started the trouble.

PR #89, opened 2016-05-23
README.md: Rearrange badges

zmwangx:
- Upstream release should precede downstream packager releases;
- License which is intrinsic to the project should precede build status which is to a large extent extrinsic.


jarun:
Thank you!

PR #90, opened 2016-05-23
Legacy cleanup

zmwangx:
Get rid of Python 2.7 and 3.2 compat stuff, and fix some code smells. See commit message for details.

I also included a trivial fix (the last commit) that removes the extra empty line on stderr in noninteractive mode, which needs to be based on this branch, and which is too minor to deserve a separate PR.


zmwangx:
CI failure can be ignored — just unusual activity.

jarun:
Thank you!

PR #92, opened 2016-05-24
Improve quit_program and read_next_command

zmwangx:
See commit messages for details.

No user facing change.

### DIFF
--- a/googler
+++ b/googler
@@ -693,7 +693,7 @@ class Result:
         # Open the URL in a web browser if option -j was specified.
         if lucky:
             self.open()
-            quit_program(conn)
+            quit_program()
 
         index_url_map[str(index)] = url
         self.print_title_and_url(index, title, url)
@@ -836,11 +836,12 @@ def new_connection(domain=None):
     return http.client.HTTPSConnection(domain if domain else server, timeout=45)
 
 
-def quit_program(conn):
-    """Closes a connection and quits the program"""
+def quit_program(exitcode=1):
+    """Close the connection and quits the program."""
 
+    global conn
     conn.close()
-    sys.exit(1)
+    sys.exit(exitcode)
 
 
 def fetch_results():
@@ -863,7 +864,7 @@ def fetch_results():
             debugp("Redirected URL [%s]" % url)
             if url.find("sorry/IndexRedirect?") >= 0:
                 error("Connection blocked due to unusual activity.")
-                quit_program(conn)
+                quit_program()
 
             conn.close()
             mark = url.find("/search")
@@ -877,17 +878,17 @@ def fetch_results():
                 resp = google_get(conn, url)
             except Exception as e:
                 debugp("Exception: %s" % e)
-                quit_program(conn)
+                quit_program()
 
             if resp.status != 200:
                 # Failed connecting to redirected server too!
                 error("First redirection failed with HTTP %d: %s" %
                       (resp.status, resp.reason))
-                quit_program(conn)
+                quit_program()
         else:
             # The server responded with an error.
             error("HTTP %d: %s" % (resp.status, resp.reason))
-            quit_program(conn)
+            quit_program()
 
     # Parse the HTML document and print the results.
     parser = GoogleParser()
--- a/googler
+++ b/googler
@@ -925,10 +925,11 @@ def read_next_command():
     global colorize, colors
 
     message = "googler (? for help)"
-    if colorize:
-        return input(colors.prompt + message + colors.reset + ' ')
-    else:
-        return input("%s: " % message)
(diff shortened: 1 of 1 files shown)


googler
```diff
@@ -836,11 +836,12 @@ def new_connection(domain=None):
     return http.client.HTTPSConnection(domain if domain else server, timeout=45)
 
 
-def quit_program(conn):
-    """Closes a connection and quits the program"""
+def quit_program(exitcode=1):
+    """Close the connection and quits the program."""
 
+    global conn
```
jarun: I wanted to avoid this. In case we ever open a new connection. Let's keep this a separate unit.

zmwangx: I think we should only have one open connection at any time. The old connection should be closed before the new one is opened. The old `quit_program(conn)` also closes one connection only.

jarun: `quit_program()` closes the connection passed to it.

`new_connection()` is tied to any global today.

> I think we should only have one open connection at any time.

That is true but we need not use the global `conn` forever. I'm thinking of supporting program options at omniprompt in future. Would add great flexibility.

zmwangx: As I mentioned `fetch_results()` use the global `conn` too, so I don't see why `quit_program()` can't use it. We can make `new_connection()` handle the closure of the old one (if not `None`) and assign to the global too. I'm fine with either.

The main reason I decided to use the global is that `quit_program` really has little to do with an arbitrary given connection. Since `quit_program` is the final function doesn't return, it should work directly with global states, not some arbitrary parameter. Whether the global states are kept in global variables, or we write a class to keep track of global states and make `quit_program` and such instance methods, I don't care much at this point. (But that's certainly a good thing for unit testing purposes.)

Also, I was trying to handle ^D in `read_next_command`, which means embedding `quit_program`. We can't pass an arbitrary `conn` into `read_next_command`, that would look ridiculous.

> I'm thinking of supporting program options at omniprompt in future. Would add great flexibility.

Greater flexibility, sure. But there's still one global connection. I don't see how that would change.

jarun:
Agreed.

jarun:
> Also, I was trying to handle ^D in read_next_command

Moving that piece of code would be nice.

jarun:
> Moving that piece of code would be nice.

Sorry, I wanted to mean the double `Enter` handling.

zmwangx:
> Sorry, I wanted to mean the double Enter handling.

```diff
From 193facc3936cfac0569b0344dba5702908439152 Mon Sep 17 00:00:00 2001
From: Zhiming Wang <zmwangx@gmail.com>
Date: Tue, 24 May 2016 00:05:21 -0700
Subject: [PATCH] googler: read_next_command: Incorporate double-enter-to-quit

---
 googler | 18 +++++++++---------
 1 file changed, 9 insertions(+), 9 deletions(-)

diff --git a/googler b/googler
index 7ad0153..11a03bd 100755
--- a/googler
+++ b/googler
@@ -926,9 +926,15 @@ def read_next_command():
 
     message = "googler (? for help)"
     prompt = (colors.prompt + message + colors.reset + " ") if colorize else (message + ": ")
-    try:
-        return input(prompt)
-    except EOFError:
+    for _ in range(2):
+        try:
+            command = input(prompt)
+        except EOFError:
+            quit_program(exitcode=0)
+        if command:
+            return command
+    else:
+        # Double Enter
         quit_program(exitcode=0)
 
 
@@ -1215,12 +1221,6 @@ while True:
 
     oldstart = start
     nav = read_next_command()
-    if not nav:
-        nav = read_next_command()
-        if not nav:
-            # Two consecutive enters
-            quit_program(exitcode=0)
-
     nav = CONSECUTIVE_WHITESPACE.sub(" ", nav).strip()
     if nav == "n":
         if len(results) == 0:
-- 
2.8.3

```

Control flow looks kinda ridiculous though.

By the way, we're mixing single quotes and double quotes in the source code. I think most if not all single quotes were introduced by me, when I pushed big blocks of code. I think it's better to fix a style. The reason I prefer single quotes to double quotes (except `"""` — out of habit — or when there are literal single quotes) is that we are not in C, or any other language that distinguish between character literals and string literals. In Python there's no character literal at all, and single quotes are easier to type (no shift). (I guess another reason is I write more shell than C, where in shell I would use single quotes to avoid unwanted escapes and such unless I need expansions. But that's just a personal matter.) What would you say?

jarun:
> Control flow looks kinda ridiculous though.

Yes, looks very odd. We'll skip it.

> we're mixing single quotes and double quotes in the source code.

I know. Single quotes it should be as I don't have any preference myself. Just the habit of C.

zmwangx:
> Yes, looks very odd.

Or repeating the same exception handling block twice. I want to avoid it (part of the reason why I moved it into the function).

> Single quotes it should be 

#93.

PR #93, opened 2016-05-24
googler: Use single quotes consistently

zmwangx:
Replace double quotes with single quotes.

See https://github.com/jarun/googler/pull/92#issuecomment-221184312 and https://github.com/jarun/googler/pull/92#issuecomment-221194202.

Let's try to consistently use single quotes in the future so that there's no need to choose, especially for new contributors (if any) who want to follow existing styles.

### DIFF
--- a/googler
+++ b/googler
@@ -82,7 +82,7 @@ exact = False               # If True, disable automatic spelling correction
 insite = None               # Google search a specific site
 output_format = 0           # 0: Regular (normal) output, 1: Json output
 noninteractive = False      # Non-interactive mode
-server = "www.google.com"   # Default server to connect to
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

jarun:
I guess we introduced an issue: https://travis-ci.org/jarun/googler/jobs/132489980

zmwangx:
Probably not in this commit, because nothing in `GoogleParser` was changed. But we'll see.

zmwangx:
I think it is just malformed HTML. In the HTML response, there are 444 open <div> tags and 445 close </div> tags. I haven't located where exactly is wrong yet.

zmwangx:
#94

PR #94, opened 2016-05-24
googler: Handle malformed HTML

zmwangx:
Sometimes the response HTML could be malformed, with more closing tags than opening tags for a certain tag type. In this case there's nothing we could do other than ignoring it and assuming a None annotation.

Example:

    > googler -d -c nl -l en rupees

A sample response: https://goo.gl/J4QNnP

It has 444 opening `<div>` tags yet 445 closing `</div>` tags:

    > grep -o '<div' googler-response-uw83ptz1 | wc -l
    444
    > grep -o '</div>' googler-response-uw83ptz1 | wc -l
    445


jarun:
Thanks!

PR #95, opened 2016-05-24
Revert install.comp and uninstall.comp targets in Makefile

zmwangx:
My points have been made clear in https://github.com/jarun/googler/commit/ebad612ee553dfef7a0e66eb34179e8c069ce503 and https://github.com/jarun/googler/pull/88#discussion_r64467089. To recap:

1. ca311b31440746151ebca0763c9c3c4ba4924225 was backed by [GNU Coding Standards on `sysconfdir`](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) and [FHS](https://www.linuxbase.org/betaspecs/fhs/fhs/ch04s09.html).

2. One should not assume user shell's installation location, and should not assume a site install. Completion scripts should be kept relative to the program's installation location, not guessed shell locations (in particular, there's no standard location for bash completion scripts). `PREFIX` should be respected, and the ability to perform a user-local install or multiple installs on the same site should not be taken away.

Of course these points are somewhat subjective. But these are tried and true at least for me. I detest installation scripts that don't respect `PREFIX`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,15 +89,7 @@ To remove `googler` and associated docs, run
 
 ### Shell completion
 
-Shell completion scripts for Bash, Fish and Zsh can be found in respective subdirectories of [`auto-completion/`](auto-completion), and can be installed through
-
-    $ sudo make install.comp
-
-and uninstalled through
-
-    $ sudo make uninstall.comp
-
-Again, `PREFIX` is supported. Please refer to your shell's manual for instructions on setting up your shell to use completions.
+Shell completion scripts for Bash, Fish and Zsh can be found in respective subdirectories of [`auto-completion/`](auto-completion). Please refer to your shell's manual for installation instructions.
 
 ## Installing with a package manager
 
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ and uninstalled through
 
     $ sudo make uninstall.comp
 
-`DESTDIR` is supported. Please refer to your shell's manual for instructions on completion configuration.
+Again, `PREFIX` is supported. Please refer to your shell's manual for instructions on setting up your shell to use completions.
 
 ## Installing with a package manager
 
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,8 @@ PREFIX ?= /usr/local
 BINDIR = $(DESTDIR)$(PREFIX)/bin
 MANDIR = $(DESTDIR)$(PREFIX)/share/man/man1
 DOCDIR = $(DESTDIR)$(PREFIX)/share/doc/googler
-BASHCOMPDIR = $(DESTDIR)$(PREFIX)/etc/bash_completion.d
-FISHCOMPDIR = $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d
-ZSHCOMPDIR = $(DESTDIR)$(PREFIX)/share/zsh/site-functions
 
-.PHONY: all install install.comp uninstall uninstall.comp
+.PHONY: all install uninstall
 
 all:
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -97,7 +97,7 @@ and uninstalled through
 
     $ sudo make uninstall.comp
```
jarun: With the proposed Makefile, our examples for install and uninstall should include `PREFIX=/usr`.
We can safely assume very few would be interested in reading their shell's manual to install scripts that come with instructions to install.

zmwangx: You mean, `PREFIX=/usr` only for the `.comp` targets?

> We can safely assume very few would be interested in reading their shell's manual to install scripts that come with instructions to install.

That could be true to some extent. Which is pretty sad.

README.md
```diff
@@ -97,7 +97,7 @@ and uninstalled through
...
 
-`DESTDIR` is supported. Please refer to your shell's manual for instructions on completion configuration.
+Again, `PREFIX` is supported. Please refer to your shell's manual for instructions on completion configuration.
```
jarun: Adding `PREFIX=/usr` in the example would render this line redundant. And we an also skip the `read your shell's manual` part.

zmwangx: >  And we an also skip the read your shell's manual part.

I don't think that's skippable. Bash and zsh doesn't have completion enabled out of box.

jarun: > I don't think that's skippable. Bash and zsh don't have completion enabled out of box.

OK, we'll keep the note.

jarun:
Here's another proposal to handle this. Instead of handling auto-complete files, what it we just mention standard locations for auto-completion script installation for each shell? Fish as [well documented locations](http://fishshell.com/docs/current/index.html#completion), bash too has `/etc/bash_completion.d`. How about Zsh?

zmwangx:
> Here's another proposal to handle this. 

Sorry, how's this different? What the `install.comp` target does is exactly installing to standard or semi-standard locations. (`$PREFIX/share/zsh/site-functions` is in `$fpath` by default, `$PREFIX/share/fish/vendor_completions.d` is in `$fish_complete_path` by default, but `/etc/bash_completion.d` is not even close to being a standard — there's no `autoload` or `fpath`, which I believe are originally ksh concepts, equivalent for bash.) `install.comp` allows users to control the `PREFIX` to install to, that's all.

jarun:
> $PREFIX/share/fish/vendor_completions.d is in $fish_complete_path by default

Only if $PREFIX = /usr of /usr/share/fish/vendor_completions.d.

Can we update the README with  `PREFIX=/usr` then?

jarun:
> Can we update the README with  `PREFIX=/usr` then?

Understood from your latest update we can't. I say we skip bothering about the script installation altogether.

zmwangx:
> Only if $PREFIX = /usr of /usr/share/fish/vendor_completions.d for fish.

Well, no, `PREFIX` is the prefix where `fish` is installed into. If you install `fish` into `/usr/local` then it will be `/usr/local/share/fish/vendor_completions.d` instead of `/usr/share/fish/vendor_completions.d`, and if you install `fish` into `/usr/local/Cellar/fish/2.3.0` then it will be `/usr/local/Cellar/fish/2.3.0/share/fish/vendor_completions.d`.

> Or does fish install to /usr/local on OS X as well?

Fish uses autotools. It completely depends on what `--prefix` you pass during the `configure` stage.

> I say we skip bothering about the comp script installation altogether.

I'm fine with that too, given the fundamental disagreements here.

zmwangx:
Updated to PR to revert `install.comp` and `uninstall.comp`.

jarun:
Thank you!

PR #96, opened 2016-05-25
Add support for unfiltering filtered results

zmwangx:
It's easy to implement, and I think it's still a worthy addition.

---

Sometimes Google omits similar (more like duplicate) result entries. Whenever this happens, a notice is displayed in `p#ofr` that reads like

> In order to show you the most relevant results, we have omitted some entries very similar to the 7 already displayed. If you like, you can repeat the search with the omitted results included.

The way to unfilter is simple. Just add `&filter=0` to the query string.

In this commit, we watch for `p#ofr`, and when detected, we set the `filtered` property of the parser. Then after showing all results, we display a notice and enable a contextual command `unfilter` in the prompt. Since it is rare and contextual, I think it is okay to give it a long name.

Sample query: `"why oh-my-zsh is completely broken"` (quotes are part of the query).

Demo:

[asciicast](https://asciinema.org/a/9afh4ut5yzus5g3l1eyax0d8t)

### DIFF
--- a/googler
+++ b/googler
@@ -310,6 +310,10 @@ class GoogleParser(html.parser.HTMLParser):
     #     abstract text
     #   </div>            <!-- stop_populating_textbuf, pop to sitelink abstract -->
     #
+    # - Sometimes Google omits similar (more like duplicate) result
+    #   entries. Whenever this happens there will be a notice in p#ofr. The way
+    #   to unfilter is to simply add '&filter=0' to the query string.
+    #
     #
     # Google News
     #
@@ -362,6 +366,7 @@ class GoogleParser(html.parser.HTMLParser):
     def __init__(self):
         html.parser.HTMLParser.__init__(self)
 
+        self.filtered = False
         self.results = []
 
         self.index = 0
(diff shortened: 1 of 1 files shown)


zmwangx:
Oh, by the way the effect of `filter=0` doesn't carry over when you start a new query, since we rebuild the URL in that case. Which is good.

jarun:
How do we switch back to (toggle) filtered results?

Also, a little unrelated, in asciinema how to you control the timing of the `snapshot`?

zmwangx:
> How do we switch back to (toggle) filtered results?

As the implementation currently stands there's no way. But is that really necessary? You can't toggle filtering back in Google's web interface either (I guess there's browser's back button, but that's not part of Google's interface per se). Just start a new query.

We also don't support toggling options inline. At least not yet.

> Also, a little unrelated, in asciinema how to you control the timing of the snapshot?

screen shot 2016-05-25 at 8 05 18 pm

jarun:
> But is that really necessary?

Nope. I didn't go through the change yet. Just your comments. Thought this piece was missing.

Thanks for the asciinema tip :)

jarun:
:+1: Thank you!

PR #99, opened 2016-05-30
OO refresh

zmwangx:
This is a huge WIP, intended to group and modularize `googler`'s various functional units, hence lifting googler from its quick-and-dirty origin. (Starting out quick-and-dirty is totally fine, but at some point we need to rethink the structure in order to not dissolve into a pile of unmaintainable mess when we add more features. That's just my opinion.)

The guiding principle here is to break down big functional units and dissociate state info so that every small blob of code should make as much sense as possible locally, instead of relying on memorizing the effect of another blob that is two hundred lines away. (Bad example: `url = url.replace('start=%d&' % oldstart, 'start=%d&' % start, 1)`, which relies on storing a snapshot of `start` in `oldstart` many lines before actually using it; this sort of disconnectedness could be a source of bugs when part of the program is updated.) When we do need persistent state info, they should be packaged into well-defined and well-documented OO interfaces, complete with relevant state manipulation helpers.

A brief and incomplete checklist:

- [x] `GooglerArgumentParser`: absorb type guards;
- [x] `GoogleUrl`: URL constructor;
- [x] `GoogleConnection`: handle connection opening/renewal and page fetching (including redirection) (should absorb `google_get`, `new_connection` and a big chunk of `fetch_results`);
- [x] `GooglerCmd`: OO command line interpreter and executioner (model on [`cmd.Cmd`](https://docs.python.org/3/library/cmd.html)?).

Classes should be global-free (except constants and `logger`):

- [x] `GoogleUrl`
- [x] `GoogleConnection`
- [x] `GoogleParser`
- [x] `Result` — `lucky` should be dealt with outside `Result`; `colors`, `columns` etc. that affect the output should be class variables; <s>`json` should be dealt with inside `Result`</s>; oh, and think of a better name for `urlindex` (`urltable` maybe?)...
- [x] `GooglerCmd`
- [x] `GooglerArgumentParser`

I'm busy recently so it will take a while.

The finished parts could be reviewed and/or regression-tested and I'd like to have feedback.

### DIFF
--- a/googler
+++ b/googler
@@ -89,7 +89,6 @@ lucky = False               # If True, opens the first URL in browser (option --
 
 # Cosmetic options
 columns = None              # Terminal width
-colorize = True             # If True, colorizes the output (option --nocolor)
 colors = None               # Colors object, set after reading color settings (option --colors)
 
 
@@ -1201,11 +1200,11 @@ class Result(object):
 
     @staticmethod
     def print_title_and_url(index, title, url, indent=0):
-        global colorize, colors
+        global colors
         # Pad index and url with `indent` number of spaces
         index = ' ' * indent + str(index)
         url = ' ' * indent + url
-        if colorize:
+        if colors:
             print(colors.index + index + colors.reset, end='')
             print(' ' + colors.title + title + colors.reset)
             print(colors.url + url + colors.reset)
@@ -1214,14 +1213,14 @@ class Result(object):
 
     @staticmethod
     def print_metadata_and_abstract(abstract, metadata=None, indent=0):
-        global colorize, colors, columns
+        global colors, columns
         if metadata:
-            if colorize:
+            if colors:
                 print(colors.metadata + metadata + colors.reset)
             else:
                 print(metadata)
 
-        if colorize:
+        if colors:
             print(colors.abstract, end='')
         if columns > indent + 1:
             # Try to fill to columns
@@ -1231,7 +1230,7 @@ class Result(object):
             print('')
         else:
             print('%s\n' % abstract.replace('\n', ' '))
-        if colorize:
+        if colors:
             print(colors.reset, end='')
 
     def print_entry(self):
@@ -1431,10 +1430,10 @@ def fetch_results():
 def read_next_command():
     """Show omniprompt and read user command."""
 
-    global colorize, colors
+    global colors
 
     message = 'googler (? for help)'
-    prompt = (colors.prompt + message + colors.reset + ' ') if colorize else (message + ': ')
+    prompt = (colors.prompt + message + colors.reset + ' ') if colors else (message + ': ')
     try:
         return input(prompt)
     except EOFError:
@@ -1517,8 +1516,10 @@ if len(sys.argv) < 2:
 
 # Set global variables
 opts = argparser.parse_args()
-colorize = opts.colorize
-colors = Colors(*[COLORMAP[c] for c in opts.colorstr], reset=COLORMAP['x'])
+if opts.colorize:
+    colors = Colors(*[COLORMAP[c] for c in opts.colorstr], reset=COLORMAP['x'])
+else:
+    colors = None
 lucky = opts.lucky
 if opts.debug:
     logger.setLevel(logging.DEBUG)
--- a/googler
+++ b/googler
@@ -21,6 +21,7 @@ from __future__ import print_function
 import argparse
 import atexit
 import collections
+import functools
 import gzip
 import http.client
 import html.entities
@@ -72,23 +73,6 @@ COLORMAP = {k: '\x1b[%sm' % v for k, v in {
 # Disguise as Microsoft Edge
 USER_AGENT = ('Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 '
               '(KHTML, like Gecko) Chrome/42.0.2311.135 Safari/537.36 Edge/12.10240')
-
-
-# Global variables
-
-# State carriers
-conn = None                 # The global connection
-google_url = None           # The URL constructor
-opts = None                 # User options
-nav = 'n'                   # User command register
-
-# Output/User interaction options
-noninteractive = False      # Non-interactive mode (option --no-prompt)
-output_format = 0           # 0: Regular (normal) output, 1: JSON output (option --json)
-lucky = False               # If True, opens the first URL in browser (option --lucky)
-
-# Cosmetic options
-colors = None               # Colors object, set after reading color settings (option --colors)
 
 
 # Global helper functions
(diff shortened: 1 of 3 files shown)


zmwangx:
By the way, this is laying groundwork for both "Option-related improvements" and "Refactor and write unit tests" on the todo list.

jarun:
> I'm busy recently so it will take a while.

Please take your time. This is massive improvement over what we have today.

zmwangx:
Pushed another commit, implementing `GoogleConnection`. That wraps up this week's work.

jarun:
:+1: get some rest.

zmwangx:
> 👍 <s>get some rest.</s> Pick up another piece of work.

FTFY 😛

zmwangx:
Seems that Google raised their bar of suspicious activity, or we have some unknown competitor testing Google search on Travis at the same time...

jarun:
> Seems that Google raised their bar of suspicious activity

We can experiment with the cookie stuff our user share on Gitter.

zmwangx:
> We can experiment with the cookie stuff our user share on Gitter.

You mean grab a cookie from the browser?

1. I don't think spamming hundreds requests with cookies is not suspicious from Google's point of view; 
2. Even if it does prevent Google's block, it only works for so long before it expires;
3. I would be hesitant to make my cookies (signed in) public, because that may open up a channel for session hijacking attacks against my Google account.

It doesn't hurt to try though.

jarun:
No no... This one: https://docs.python.org/3.0/library/http.cookiejar.html

Now we are all-Python3. No double maintenance.
If we do it, we handle ourselves.

jarun:
> I don't think spamming hundreds requests with cookies is not suspicious from Google's point of view

We can check this by firing 100 requests from the browser. I can't do it right now.

zmwangx:
Oh, you mean just keeping the cookies? Not sure that would help. I would say keeping the cookies would be *more* suspicious, because you're clearly spamming from the same session, rather than having multiple people searching from the same IP.

What I had in mind was cookies from a signed-in session, because I've been never greeted by a ReCAPTCHA when I'm signed in, but occasionally do when I search from Incognito/Private, especially when using my PIA VPN.

zmwangx:
> What I had in mind was cookies from a signed-in session

Obviously we don't want to introduce signing in.

Anyway, it's not a big deal. Also, my plans for testing would render this spamming a thing of the past.

jarun:
> Not sure that would help.

This is what I wanna test, from a fresh, cookie-cleared browser session.

> Obviously we don't want to introduce signing in.

We can't. Google has added restrictions recently. Many third-party clients can't anymore.

zmwangx:
> We can't. 

Plus that hardly helps our users. Only helps with automated testing.

zmwangx:
> This is what I wanna test, from a fresh, cookie-cleared browser session.

And I just tried.

```bash
for i in {1..500}; do echo -n $'\r'$i; chrome-cli open 'https://google.com/search?q=hello' >/dev/null; sleep 1; chrome-cli close; done
```

All pages are opened in a fresh Incognito session. Got ReCAPTCHA'ed on the ~100th request.

jarun:
> Got ReCAPTCHA'ed on the ~100th request.

LoL!!! Doesn't help then.

zmwangx:
@jarun This branch is ready for review.

IMO the first thing to do is to review the commit messages of each commit (which are very detailed, especially on potentially controversial things) and let me know which points you're unhappy with.

zmwangx:
By the way, stats show an increase of ~500 lines, which is kind of scary, but according to `ohcount`, on master (08b39e540d82cf190a35408f15b025f09fe9de4f) we have

```
Language          Files       Code    Comment  Comment %      Blank      Total
----------------  -----  ---------  ---------  ---------  ---------  ---------
python                1        759        365      32.5%        180       1304
```

and on `oo-refresh` (ca33fda1b93a2641e424a893ac60f23f2710fa14) we have

```
Language          Files       Code    Comment  Comment %      Blank      Total
----------------  -----  ---------  ---------  ---------  ---------  ---------
python                1        907        630      41.0%        273       1810
```

So more comprehensive documentation as well as more blank lines from more `def`s account for most of the increase.

zmwangx:
A word on performance: just from noninteractive testing, there's no noticeable difference between `master` and `oo-refresh`. Of course, I can't test for thousands of runs, so the fluctuations are pretty big, and either one could be faster in my several batch-of-10 runs due to fluctuations.

jarun:
How about a separate text doc for API documentation?

zmwangx:
> How about a separate text doc for API documentation? 

1. Docstrings are the convention;
2. Docstrings of methods and classes can't really be separated from the source code, or they'll make no sense;
3. Separate documentation gets outdated (and worse, goes unnoticed) easily;
4. The only large block of docs here is for the parser.

We're at the point (if this PR is merged) where I can generate API docs in HTML with Sphinx if you'd like to have that.

jarun:
> Separate documentation gets outdated (and worse, goes unnoticed) easily 

Something I'm afraid of too. But in case it gets in the way of performance, for stabler code and fundamental design we can do this.

> We're at the point (if this PR is merged) where I can generate API docs in HTML with Sphinx if you'ld like to have that.

Nope. Not before someone wants to use it as a library.

zmwangx:
> But in case it gets in the way of performance

? Docstrings gets in the way of performance? No.

jarun:
> Docstrings gets in the way of performance? No.

Right. Sorry, just reading a few lines won't hurt.

zmwangx:
@jarun In case I didn't make it clear, this branch is ready for review. It's totally fine if you don't have time to review (it will be a pretty significant undertaking).

I can pull in #104 if you give the nod to that one, which will reduce the review burden here because the entire `GoogleConnection` class will be eliminated.

jarun:
> I can pull in #104 if you give the nod to that one

That would be great!

googler
```diff
@@ -19,16 +19,21 @@
 
 from __future__ import print_function
```
jarun: Do we need this anymore now that we are all Python3?

googler
```diff
@@ -740,313 +1324,316 @@ class Result:
...
+        self._opts = opts
 
-def message(msg, level='INFO', ansi=None):
-    """Print message, led with [%(level)s] where level is a string, to
-    stderr.  If ansi is not None and the global variable colorize is True,
-    then print ansi at the beginning, and close with \x1b[0m.
-    """
+        self._google_url = GoogleUrl(opts)
+        self._conn = GoogleConnection(self._google_url.hostname)
+        atexit.register(self._conn.close)
```
jarun: Does this mean we are closing the connection after 10 fetches (each page)?
Also, I don't think importing a library (atexit) for a single function is required.

googler
```diff
@@ -19,16 +19,21 @@
 
 from __future__ import print_function
 import argparse
+import atexit
 import collections
+import functools
```
jarun: Really necessary?

googler
```diff
@@ -19,16 +19,21 @@
...
 import gzip
 import http.client
 import html.entities
 import html.parser
 import io
+import json
+import logging
```
jarun: Please late import this only if logging is requested.

googler
```diff
@@ -19,16 +19,21 @@
...
 import os
 import re
 import signal
 import sys
+import tempfile
```
jarun: Please late import.

jarun:
Many of the comments are related to imports. The general problem I see with Python is - you import 3/4 libs, you add 150-180ms delay. I want to avoid this.

googler
zmwangx: > Does this mean we are closing the connection after 10 fetches (each page)?

No. This is an at exit handler. Only called when program exits. And we only have one instance of `GoogleConnection` throughout the program, updated through `new_connection` and `renew_connection`.

> Also, I don't think importing a library (atexit) for a single function is required.

Or you need to have global variables and another function, instead of the standard `sys.exit`. Like the old `quit_program`. The more cross-referencing you have, the less readable and maintainable the program becomes. Therefore, registering the at exit handler at the place where it is defined is really good.

Also, most PSL module imports are extremely fast. The time to import one module is usually negligible compared to the startup time of the interpreter myself. (I believe there are exceptions, e.g., `re`, which is kind of expensive.)

googler
```diff
@@ -740,313 +1324,316 @@ class Result:
...
 
-# Messaging wrappers
+        self._opts = opts
 
-def message(msg, level='INFO', ansi=None):
-    """Print message, led with [%(level)s] where level is a string, to
-    stderr.  If ansi is not None and the global variable colorize is True,
-    then print ansi at the beginning, and close with \x1b[0m.
-    """
+        self._google_url = GoogleUrl(opts)
```
jarun: Is `GoogleUrl` initialized per new search or per instance of `googler`?

zmwangx:
> The general problem I see with Python is - you import 3/4 libs, you add 150-180ms delay.

Benchmark/citation needed. I just tested all 22 imports combined, and they take a total of 60–70ms. The interpreter startup time is about 35ms. So the time of one import is almost negligible compared to interpreter startup time. Plus networking delay is much longer that.

I guess it could take somewhat longer on magnetic storage, but I find 50ms for a single module hard to believe unless I see stats.

googler
zmwangx: Logging is always required, because errors and warnings are emitted through the logger.

googler
zmwangx: In decorators, if you don't wrap the decorated methods, you lose their signatures and docstrings. I'm not wrapping the `GoogleParser` methods because I consider it an opaque class (that's why the documentation there is in comments, not docstrings). But for `GooglerCmd`, I'm exposing the API, so the methods should be wrapped.

googler
zmwangx: I'll step down by importing `tempfile` and `json` before they're called. Importing at argparsing time really is pretty bad, it's too hard to keep track. I just checked, even pylint can't catch unused imports at random locations within functions.

googler
zmwangx: Per new query. If you don't change the keywords, it's just updated. If you change the keywords, a new instance is created with the given options. The reason is to get rid of ephemeral changes that shouldn't be kept between queries; currently, the only thing that belongs to that category is the `filter` query.

`GoogleUrl` is fast. There's not a single expensive operation in it.

jarun:
Done the review.

zmwangx:
If you really want to improve startup time, `webbrowser` can be moved to where it's used to. I don't like this style as you already know, but since we've adopted this style already, there's little reason to keep `webbrowser` at the top.

zmwangx:
Lazy import done. Rebasing was kind of a nightmare.

Any other comments or replies to my comments?

zmwangx:
Rebasing this on master won't be fun, because that cyan change would result in amending almost every single commit. If you're happy with this branch, I'll do a manual merge instead, which is easy.

googler
jarun: :+1:

googler
jarun: :+1:

googler
jarun: :+1:

jarun:
> Lazy import done. Rebasing was kind of a nightmare.

I guessed it but I really crave for speed.

> I'll do a manual merge instead, which is easy.

Go as you like :)

zmwangx:
> but I really crave for speed.

Needed to rebase anyway. No problem.

zmwangx:
Merged.

jarun:
:+1:

jarun:
I'm planning to revert this commit. Earlier I was thinking of supporting all options at the prompt and this commit would help with that but I have dumped the plan.

This commit takes re-directions to a new level. Just to figure out where program args (opts) are being parsed I had to go though 4 functions where opts is being passed around!

Let's keeps things simpler. So much for object-oriented.

jarun:
OK, that came in with GooglerCmd.

jarun:
And we couldn't live without re???

zmwangx:
> Earlier I was thinking of supporting all options at the prompt and this commit would help with that but I have dumped the plan.

I think it really makes sense to support that. But if you don't want to, we can dump `GooglerCmd`. Other classes can stay.

> This commit takes re-directions to a new level.

Eliminating side effects come with a price. Otherwise something like `devbin/parse` would not work (or it could work, but you have `global` everywhere, which would be insane).

> Just to figure out where program args (opts) are being parsed I had to go though 4 functions where opts is being passed around!

Dynamic URL construction makes sense even if you don't want to support dynamically updating options during the execution. Previously you need to understand all URL manipulation code scattered around just to make sure your redirection code won't break anything (and you actually can't). Now the complexity has been contained, and you don't have to understand anything about URL construction unless you need to.

Plus, tracebacks in the OO setting are arguably more helpful. You have the entire detailed callstack, while you could also quickly isolate the issue, rather than worrying about states.

zmwangx:
In short, if you confirm that the plan to support dynamic options has been dropped, then I'll go ahead and try to eliminate `GooglerCmd`, because that's the only class that's adding complexity (in addition to wrapping a few globals) if we don't want to actually achieve something with it.

zmwangx:
Forgot to reply to `re` comment.

Using `re` to replace consecutive Unicode whitespace
* Results in cleaner queries;
* Makes it more flexible for the users who might type in whitespace characters other than the space (e.g., when you use a Chinese input method).

But it is by no means necessary.

jarun:
> Other classes can stay.

For generating a URL (which could have been done easily with string manipulation and class variables) we have used dictionary, multiple levels of API calls to the extent that it becomes boring to understand the code... it's everywhere! This has now become a Python-internal DS usage demonstration rather than a utility one would want to go through and contribute to. Why couldn't we keep things simpler?

> Dynamic URL construction makes sense even if you don't want to support dynamically updating options during the execution.

I'll get back to this but if this is what it takes to re-form and generate a URL, I'm not sure I'll carry it along.

`googler` doesn't have to be that complex a utility. We don't want to turn it into a full-fledged enterprise database management system loaded with cannons.

zmwangx:
> we have used dictionary

Because URL query is a dictionary. It's the natural data structure to use.

> multiple levels of API calls

To construct, you call the constructor with options; to update, you call `update`; to get the URL, you access the `url` property, or `full()` (same thing, just for clarity in certain use cases) or `relative()` depending on use case. Navigation is handled too. What is multi-leveled? If you don't need dynamic manipulation of options, just ignore some of the methods. You don't have to use all methods, and existence of a few possibly extraneous methods doesn't make the class more complex.

> googler doesn't have to be that complex a utility.

Number of lines isn't equal to complexity. Complexity is nothing if it's contained and split up.

PR #109, opened 2016-06-06
Baby step: Allow no keywords on the command line, and introduce -h, --help option

zmwangx:
This is point 4 of https://github.com/jarun/googler/issues/87#issuecomment-221962173.

---

All the machinery is almost there, just need to flip the switch.

Instead of printing the help text when there's is no argument, we now support the standard `-h, --help` option, and a command line with no argument leads us into the REPL immediately.

Transcript of a sample session:

    > googler -C -w github.com -n 3
    Please initiate a query.
    googler (? for help): f
    Initiate a query first.
    googler (? for help): n
    Initiate a query first.
    googler (? for help): o
    Initiate a query first.
    googler (? for help): googler

     1 GitHub - jarun/googler: Google from the command-line
    Google from the command-line. Contribute to googler development by
    creating an account on GitHub.

     2 googler/googler.1 at master · jarun/googler · GitHub
    https://github.com/jarun/googler/blob/master/googler.1
    Google from the command-line. Contribute to googler development by
    creating an account on GitHub.

     3 googler/README.md at master · jarun/googler · GitHub
    https://github.com/jarun/googler/blob/master/README.md
    Google from the command-line. Contribute to googler development by
    creating an account on GitHub.

    googler (? for help): q

Also fixed a bug in `GooglerArgumentParser` where file is not being defaulted to `sys.stderr`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ Please substitute `$version` with the appropriate package version.
       KEYWORD               search keywords
 
     optional arguments:
+      -h, --help            show this help message and exit
       -s N, --start N       start at the Nth result
       -n N, --count N       show N results (default 10)
       -N, --news            show results from news section
--- a/auto-completion/bash/googler-completion.bash
+++ b/auto-completion/bash/googler-completion.bash
@@ -10,7 +10,7 @@ _googler () {
     local IFS=$' \n'
     local cur=$2 prev=$3
     local -a opts opts_with_args
-    opts=(-c --tld -C --nocolor -d --debug -j --first --lucky --json -l --lang
+    opts=(-c --tld -C --nocolor -d --debug -h --help -j --first --lucky --json -l --lang
           -n --count -N --news --np --noprompt -s --start -t --time -w --site -x --exact)
     opts_with_arg=(-c --tld --colors -l --lang -n --count -s --start -t --time -w --site)
 
--- a/auto-completion/fish/googler.fish
+++ b/auto-completion/fish/googler.fish
@@ -4,6 +4,7 @@
 # Author:
 #   Arun Prakash Jana <engineerarun@gmail.com>
 #
+complete -c googler -s h -l help           --description 'show help text and exit'
 complete -c googler -s s -l start  -r      --description 'start at the Nth result'
 complete -c googler -s n -l count  -r      --description 'show specified number of results (default 10)'
 complete -c googler -s N -l news           --description 'show results from news section'
--- a/auto-completion/zsh/_googler
+++ b/auto-completion/zsh/_googler
@@ -9,6 +9,7 @@
 setopt localoptions noshwordsplit noksharrays
 local -a args
 args=(
+    '(- : *)'{-h,--help}'[show help text and exit]'
     '(-c --tld)'{-c,--tld}'[country-specific search with top-level domain]:top level domain without dot'
     '(-C --nocolor --colors)'{-C,--nocolor}'[disable color output]'
     '(-C --nocolor)--colors[set output colors]:six-letter string'
--- a/googler
+++ b/googler
@@ -1617,6 +1617,9 @@ class GooglerCmd(object):
         """Run REPL."""
         if self.keywords:
             self.fetch_and_display()
+        else:
+            printerr('Please initiate a query.')
+
         while True:
             self.read_next_command()
             # TODO: Automatic dispatcher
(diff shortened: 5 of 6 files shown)


zmwangx:
Damn, really hope we have some sort auto-generation mechanism for the man page.

jarun:
[Grapse](https://tuxdiary.com/2016/01/22/grapse/)

jarun:
Thank you!

zmwangx:
> Grapse helps to some extent.

The demo seems to be in `mdoc` with a quick glance. I suppose it has a `groff_man` only mode?

zmwangx:
It doesn't seem to work?

screen shot 2016-06-05 at 8 23 07 pm
screen shot 2016-06-05 at 8 23 10 pm

zmwangx:
Or should we switch to mdoc? AFAIK mdoc is certainly more pleasant than `groff_man`, and arguably better documented, IIRC.

zmwangx:
Actually, although I had no luck with http://www.roperzh.com/grapse/, live preview for man pages is a very good idea. I'll hack up something this weekend. The idea is very simple, and is easily done with a bit of Python/Shell and a bit of Javascript:

- `groff_man` to generate HTML;
- `webbrowser.open` to open HTML, except that we append a `<script>` tag for auto refreshing;
- Have a daemon watch for source file changes, regen HTML on change.

jarun:
> It doesn't seem to work?

Hmmm... That's a problem. I never tested with `googler`.

I gifted them their [first bug](https://github.com/roperzh/grapse/issues/1) and we'll wait for a while on this :).

zmwangx:
They don't support `groff`, that's all.

jarun:
Ahh, just saw their readme.

jarun:
This should work for you (OS X only): http://sveinbjorn.org/mandrake

jarun:
I'm going through mdoc.

zmwangx:
> http://sveinbjorn.org/mandrake

Looks nice, but that also seems to be a mdoc editor. At least it won't render our man page.

zmwangx:
`groff -man -Thtml` output is crap, but `man`'s plain text output is very easy to parse, so it's easy to convert it to HTML. I'll do the previewer when I have time.

zmwangx:
Just to give you a sense of what I could achieve with a pretty trivial parsing script:

screen shot 2016-06-05 at 9 54 38 pm

I'll call it a day, and prepare for my tomorrow's exam.

jarun:
:+1:

jarun:
I think you should create a new project and share this with the world. ;)
Many devs will be very happy.

zmwangx:
> I think you should create a new project and share this with the world. ;)

Exactly what I had in mind. But in reality it would be something trivial, like a single Python script.

jarun:
Yes, but it's better that one person writes it than everyone writing it.

jarun:
BTW, quick question because I didn't reach there yet (and donno where it is):

If we move to mdco with the command:

    man googler

work? Or we need to invoke something else?

jarun:
:+1:

jarun:
> You can check the man page of BSD tar on Linux if you have time

Will confirm in the evening.

zmwangx:
As you can see, text processing in Python... Such a shame.

But I'm not a Perl wizard, and even worse at Awk, so that's it, a poor performing piece of crap that somehow does the work (almost).

jarun:
I think in a combat it's better to use known weapons than suicide with brand new ones (without proper training).

jarun:
> It should work, but need to double check on Linux.

Just checked. Works fine.

jarun:
I found this one as well: [ronn](https://github.com/rtomayko/ronn#readme)

zmwangx:
Knew that one, not impressed. I once tried all Markdown to man (groff) converters on the market and was very disappointed by the quality (not sure if the quality of stuff have since improved). Among lightweight markup languages I only know AsciiDoc (or rather, the implementation `a2x(1)`) as being reasonably good at man pages: http://www.methods.co.nz/asciidoc/chunked/ch24.html. Plus AsciiDoc is [kinda endorsed by Linus](https://plus.google.com/+LinusTorvalds/posts/X2XVf9Q7MfV).

jarun:
asciidoc looks good too. I can try this out. Anything that makes life easier with man pages.

PR #110, opened 2016-06-06
Fix running from outside US

phihag:
Previously, running googler from Germany resulted in an error:

    [ERROR] Got HTTP 302: Found

The fix is quite simple: Upon detecting a 302 (https://www.google.com/search... -> http://www.google.de/search... ), googler already correctly redirects to the new URL. However, after the redirect request has been sent, the old request was used before when going on. Instead, use the newly sent request.

### DIFF
--- a/googler
+++ b/googler
@@ -603,6 +603,7 @@ class GoogleConnection(object):
             if 'sorry/IndexRedirect?' in redirection_url:
                 raise GoogleConnectionError('Connection blocked due to unusual activity.')
             self._redirect(redirection_url)
+            resp = self._resp
 
         if resp.status != 200:
             raise GoogleConnectionError('Got HTTP %d: %s' % (resp.status, resp.reason))


jarun:
Thank you very much for finding the bug!

Though I use it from India I didn't receive the error as I set the server to India explicitly (for faster responses). And @zmwangx is in the US.

zmwangx:
Yes, that was a blunder. Thanks.

jarun:
Unacceptable miss from my side... I have switched back to default settings.

zmwangx:
@jarun I think you just need to run with default settings when you're testing, or have multiple aliases. For instance, I have `googler` aliases to `googler --colors bjdxxy` but `goo` aliased to `googler --noprompt`. I usually use `goo`, but I use `googler` instead when I test the interactive components.

jarun:
In general I run an extensive test before a release. As you see, this is not released yet. However, I could have caught it because I use `googler` myself, had I used default settings.

zmwangx:
I think it's better to test each PR before merging. But it was my fault, I'm not trying to shift the blame.

jarun:
> I think it's better to test each PR before merging.

True, but not sure how much it would happen in practice. I tend to accept PRs to master after a regular review, go through all changes before a release, write down the test cases and then execute the tests.

As PRs are asynchronous, it would be difficult for me to allocate time and test each one individually.

jarun:
Maybe down the line we can have a PR template (like we have for issues today) where we request the list of tests executed. A second /third eye inspection might uncover more test cases.

zmwangx:
> As PRs are asynchronous

True, but I think merging could wait until proper testing. IMO software developed in the open in this day and age should strive to be stable with each commit on master.

> the list of tests executed

The test script is extremely limited for now, and the regional issue is one thing that's extremely hard to properly resolve without a group of international testers (that's why our US-based CI also didn't pick up the blunder).

jarun:
> The test script is extremely limited for now 

I meant tests run by the dev, from the script or on his own manually.

zmwangx:
> I meant tests run by the dev, from the script or on his own manually.

I'm confused. You mean `devbin/parse`? That's not the cure to everything; it's a development aid for `GoogleParser`, nothing more. Or `tests/test`? That's the one I called extremely limited.

jarun:
> I'm confused.

I'm talking of a [PR template](https://github.com/blog/2111-issue-and-pull-request-templates) where we accept PRs only if it has the list of test cases run by the contributor. That way we can review that list and point out if something else should also be tested. The list would work as an additional guideline to which areas the changes affect.

zmwangx:
> I'm talking of a PR template where we accept PRs only if it has the list of test cases run by the contributor. 

But that's the job of continuous integration...

jarun:
There are instances where it may not be possible. Like the one in this case. Just a thought.

jarun:
And there would be cases where it's not possible to know what tests should be added before the PR is raised... new features for example.

zmwangx:
If you just want to ask people to run `tests/test`, that's exactly what the CI does. Of course, more runs from different locations mean more chances of surfacing bugs, but most of the time it offers diminishing returns compared to just letting the CI do its job.

The inherent lack of translation invariance and time invariance of our utility means that a satisfactory solution is pretty much impossible. I'd say let's just stick with what we have and be more cautious in the future (note to self).

jarun:
:+1:
We'll re-look at it if we hit a second case.

PR #111, opened 2016-06-08
googler.1: Simplify the SYNOPSIS section

zmwangx:
No one ever remembers to update the synopsis, so it's already horribly outdated. Moreover, without an auto generator, listing and formatting all the options is a pain in the ass. Just leave options to the `OPTIONS` section.

### DIFF
--- a/googler.1
+++ b/googler.1
@@ -2,10 +2,7 @@
 .SH NAME
 googler \- Google from the command-line
 .SH SYNOPSIS
-.B googler
-[-s N] [-n N] [-N] [-c TLD] [-l LANG] [-x] [-C]
-.br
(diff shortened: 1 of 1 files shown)


jarun:
Thank you! We also need to add debian to the readme installation section. :)

zmwangx:
We can say a word about googler available in sid, but I'm no Debian fan so I'm not sure how to install sid packages on a stable channel (except the `debootstrap` approach which I do know, but that's installing one distro on top of another). There's http://serverfault.com/q/22414 but I don't bother to try those myself.

zmwangx:
I also have concerns about the package in its current form which I don't have time to elaborate on right now. I'll talk to you about that tomorrow on a private channel.

jarun:
sure!

jarun:
I added that.

Issue #115, opened 2016-06-13
2.5 broken on OS X: module 'socket' has no attribute 'TCP_DEFER_ACCEPT'


zmwangx:
OS X 10.11.5, Python 3.5.1, googler 2.5:

```python
> /usr/local/Cellar/googler/2.5/bin/googler --debug --noprompt Homebrew
[DEBUG] Connecting to new host www.google.com
Traceback (most recent call last):
  File "/usr/local/Cellar/googler/2.5/bin/googler", line 579, in new_connection
    self._conn.connect()
  File "/usr/local/Cellar/googler/2.5/bin/googler", line 122, in connect
    sock.setsockopt(socket.SOL_TCP, socket.TCP_DEFER_ACCEPT, 1)
AttributeError: module 'socket' has no attribute 'TCP_DEFER_ACCEPT'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/Cellar/googler/2.5/bin/googler", line 1908, in <module>
    main()
  File "/usr/local/Cellar/googler/2.5/bin/googler", line 1882, in main
    repl = GooglerCmd(opts)
  File "/usr/local/Cellar/googler/2.5/bin/googler", line 1471, in __init__
    self._conn = GoogleConnection(self._google_url.hostname, proxy=proxy)
  File "/usr/local/Cellar/googler/2.5/bin/googler", line 530, in __init__
    self.new_connection(host, port=port, timeout=timeout)
  File "/usr/local/Cellar/googler/2.5/bin/googler", line 582, in new_connection
    raise GoogleConnectionError(msg)
__main__.GoogleConnectionError: Failed to connect to www.google.com: module 'socket' has no attribute 'TCP_DEFER_ACCEPT'.
```

zmwangx:
http://linux.die.net/man/7/tcp

Apparently `TCP_DEFER_ACCEPT` is not only not POSIX, but also not available in Linux before 2.4.

jarun:
OK. I'll release 2.5.1 shortly.

jarun:
Closed by mistake.

jarun:
I think we can safely assume either none or very small section of our users are on Linux 2.4. So I'll make this change for OS X only.

I am seeing more that 100 ms gain with this option enabled.

Can you please check if https://github.com/jarun/googler/commit/99c13360d7a75f82ed4d6ee9e721f2c181fe17f8 works for you?

zmwangx:
[...]

PR #116, opened 2016-06-13
I quit

zmwangx:
Remove my name from the project. I quit.

Normally I would explain thoroughly, but I'm too tired to explain now, so I'll just be a jerk and quit without a satisfactory explanation, except for briefly pointing out that

- I don't want to put my name under this piece of software anymore (license is a factor but not the only one); 
- I'm getting tired; 
- This was not planned; kind of mood driven, but I won't regret it.

I won't even contribute fixes anymore (in the parts that I wrote), which I promised earlier. Please feel free to remove my collaborator privilege, or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,6 @@ Site specific search continues at omniprompt. Use the `g` key to run a regular G
 
 1. Copyright (C) 2008 Henri Hakkinen
 2. Copyright (C) 2015-2016 [Arun Prakash Jana](mailto:engineerarun@gmail.com)
-3. [Zhiming Wang](https://github.com/zmwangx)
 
 Special thanks to [jeremija](https://github.com/jeremija), [shaggytwodope](https://github.com/shaggytwodope) and [Narrat](https://github.com/Narrat) for their contributions and efforts in spreading `googler`.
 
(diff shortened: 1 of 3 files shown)


jarun:
3 things:

- I mentioned that I'll re-look at it.
- I was doing things in a hurry because I've to leave and I wanted to make a release for you. I know it's late there.

I think there's a miscommunication here if you think I'm not paying heed to your suggestions. Anyway, I'll make the release in the afternoon in a calm head and with the changes done the proper way.

zmwangx:
> I mentioned that I'll re-look at it.

That's not a factor.

> I was doing things in a hurry because I've to leave and I wanted to make a release for you.

Again, that's not a factor.

> if you think I'm not paying heed to your suggestions. 

That's not the case.

> mood driven

All factors are things that happened **before** #115, which I won't explain in detail.

zmwangx:
Actually, please just think of this as

- I'm busy with other (new) things in life so I don't have time for this;
- I've lost interest.

both of which are factors.

jarun:
Understood! All the best! :+1: 

Thanks a lot for you contributions. You'll remain a collaborator as I don't see a reason from my side to remove it. And you're welcome back when you have a change of heart.

zmwangx:
I do wish to remove my name though.

PR #118, opened 2016-06-17
Zsh completion: Do not mark --colors and --nocolor as mutually exclusive

zmwangx:
`--colors` is often used in aliases and `--nocolor` is supposed to be an override option most of the time, so marking them mutually exclusive only makes it hard to override.


jarun:
Thank you! :)

PR #119, opened 2016-06-24
googler: Remove from __future__ import print_function

zmwangx:
Apparently we forgot to remove the future import when dumping Python 2.7.


jarun:
Thank you!

Issue #120, opened 2016-07-18
Self upgrade for those who just want to download the frigging executable


zmwangx:
I suggest adding another completely BS-free installation method:

```sh
$ sudo curl -o /usr/local/bin/googler https://raw.githubusercontent.com/jarun/googler/v2.6/googler && sudo chmod +x /usr/local/bin/googler
```

The version number should of course be updated with each release.

To accompany this, I propose that we take a page from the book of `youtube-dl` and friends by adding a `-U, --update, --upgrade` option to do self-upgrade for the free souls that are not bound by package managers. (Truth be told, I'm not one of these free souls, but I decided to free myself after realizing just now that googler from Linuxbrew is [still stuck on v2.3](https://github.com/Linuxbrew/homebrew-core/blob/master/Formula/googler.rb) which doesn't support `--colors`...). An `--include-git` option (or something like that if anyone could come up with a better name) could also be supplied to include non-release versions from git master.

The implementation should be rather trivial through querying GitHub's API. I can take care of the implementation if this is a good idea. Cheers.

zmwangx:
By the way, we could offer a global switch, say

```python
ENABLE_SELF_UPGRADE=True
```

so that packagers can easily turn this entire feature off by substituting a single line.

jarun:
:+1:  (for both)

zmwangx:
Done in #121.

PR #121, opened 2016-07-19
Implement in-place self-upgrade mechanism

zmwangx:
Implement #120. Also corrected a few things along the way.


jarun:
Thank you!

PR #122, opened 2016-07-21
self_upgrade: Use an atexit hook for tempfile removal

zmwangx:
Using an atexit hook to remove the downloaded script when necessary means no remnant even in case of an exception.


jarun:
Thank you!

PR #123, opened 2016-07-28
Remove Linuxbrew from installation options

zmwangx:
Your call, but I'm not a big fan of Linuxbrew these days because it is regularly outdated — remember the Linuxbrew version of googler stuck on 2.3 in mid July (#120)? I still use it for many things, but for our lightweight project with no deps, there are much better options like downloading the script directly and letting it update itself.

While we're at it, I'm updating the OS X moniker per Apple's rebranding.


jarun:
Merged.

PR #127, opened 2016-08-26
tests/test: Add default values for NUM_TEST_ITERATIONS and SLEEP_DURATION

zmwangx:
https://github.com/jarun/googler/commit/1606369 broke the test script when no value is supplied for either of these two env vars.

Also document these env vars in the help message.


jarun:
Thank you!

PR #129, opened 2016-08-29
Makefile: disable-self-upgrade: Use -i.bak instead of -i for sed

zmwangx:
Unfortunately, there's a subtle difference between the -i option of GNU and BSD sed: GNU sed interprets `-i` as `-i''` (no backup), while BSD sed picks up the next argument as the backup extension. We have to offer a backup extension in place for the option to be recognized unambiguously by both GNU and BSD sed.


jarun:
Thank you!

PR #130, opened 2016-09-07
README.md: Update Asciinema screencast

zmwangx:
Got a new screencast for ya. For the record, the terminal size is 100x49.

- Now using `w3m` as `BROWSER` to avoid the awkwardness of nothing happening after entering an index;
- Self-upgrade mechanism demo at the end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #140, opened 2016-11-05
Honoring the environment variable https_proxy

snipem:
If set the environment variable https_proxy is extracted and set as
opts.proxy

PR #141, opened 2016-11-05
Properly support https_proxy

zmwangx:
Correct a few mistakes/oversights in #140:

- Env var overriding command line argument;
- Bare host:port (like localhost:8118) isn't parsed properly;
- In the man page, `https_proxy` is documented anywhere "formal" (EXAMPLES is an optional section).


jarun:
Thank you!

PR #142, opened 2016-11-11
Feature/proxy username

cosgroma:
I was having some issues getting the proxy to work when my env variable has my username and password for authentication.

I fixed that problem but broke the TLS wrapper that you have for linux tweaks. I am needing to call the HTTPSConnection directly.

Pull it in if you would like. Just offering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,7 +337,7 @@ Site specific search continues at omniprompt. Use the `g` key to run a regular G
 
         $ googler --proxy localhost:8118 google
 
-    By default the environment variable `https_proxy` is used, if defined.
+    By default the environment variable `https_proxy` is used, if defined. Supports embedded username 'username:password@proxy:port'
 
 16. More **help**:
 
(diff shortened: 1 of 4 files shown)


zmwangx:
Please don't make whitespace changes, thanks.

Also, looking at [what has actually changed](https://github.com/jarun/googler/pull/142/files?w=1), I don't think simply commenting out existing logic makes it mergeable.

If you would like to contribute proxy authentication support, please do it properly by parsing the proxy string at arg parser level, and tweak whatever wrapper as necessary to accommodate proxy authentication. Otherwise, I'm afraid you'll have to keep your patched version to yourself.

jarun:
@cosgroma option `--notweak` doesn't work for you?

cosgroma:
@zmwangx sorry about the white space, Sublime/PEP8 auto format thing... it was worse before I disabled tabs (4) -> spaces (2).

Yes, it was sloppy to comment out my changes and issue the pull, will fix.

@jarun it wasn't that ```--notweak``` didn't work, as you said, I just need to handle how the tunnel gets set up better. 

I will close this, make the changes, then resubmit.

PR #147, opened 2016-11-24
googler: fix tunneling in TLS1_2Connection

### LINKED ISSUE #146: Need Python 3.4 or plus to support https proxy in googler
#### Need Python 3.4 or plus to support https proxy in googler

According to your documentation, googler requires Python 3.3 or later. However, when I use googler with https proxy (--proxy) in Python 3.3 environment, it shows unable to connect to the http proxy server and exit.

Then I noticed that a comment in your code that "NOTE: TLS 1.2 is supported from Python 3.4", so I tried to use https proxy in Python 3.4 environment, and it worked.

I think it would better you point it out in the document. I would save a lot of time of people like me:smile::smile:.

Regards,
Eric

zmwangx:
`HTTPSConnection.connect` calls `self._tunnel` already when `self._tunnel_host` is set. (CPython 3.3.6, 3.4.5, 3.5.2, 3.6.0b4.) Ref: https://github.com/python/cpython/blob/3.3/Lib/http/client.py#L819-L824.

Fixes #146.

A quick comparison of `http/client.py` between the Py33 and Py34 branches doesn't tell me why the problem isn't seen on 3.4, though.

I still have reservations about using internals, but this is the quickest fix, for a code path that I don't care much about 😜.

### DIFF
--- a/googler
+++ b/googler
@@ -158,7 +158,6 @@ class TLS1_2Connection(HTTPSConnection):
 
         if getattr(self, '_tunnel_host', None):
             self.sock = sock
-            self._tunnel()
             HTTPSConnection.connect(self)
         elif not notweak and sys.version_info >= (3,4):
             # Use TLS 1.2


jarun:
Thank you!

PR #148, opened 2016-11-24
googler: revamp TLS 1.2 support

zmwangx:
OpenSSL has deprecated all version specific protocols. As such, version
specific protocols including PROTOCOL_TLSv1_2 have been deprecated in
Python since 3.5.3. PROTOCOL_TLS coupled with SSLContext options to
disable older protocol versions is the new way to go. See
https://docs.python.org/3.5/library/ssl.html#ssl.PROTOCOL_TLSv1_2 and
https://docs.python.org/3.5/library/ssl.html#protocol-versions.

We are now using feature test (hasattr) instead of version test which is
generally preferred.

Also, key_file and cert_file code has been removed since they're None
anyway (they are only set when key_file and cert_file are passed to
HTTPSConnection, something we don't have).

### DIFF
--- a/googler
+++ b/googler
@@ -158,13 +158,23 @@ class TLS1_2Connection(HTTPSConnection):
 
         if getattr(self, '_tunnel_host', None):
             self.sock = sock
-            HTTPSConnection.connect(self)
-        elif not notweak and sys.version_info >= (3,4):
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #153, opened 2016-12-11
Makefile: install shell wrapper with PYTHONIOENCODING=utf-8

zmwangx:
`PYTHONIOENCODING` has been a headache for us since the very beginning. Here we solve it once and for all (for Makefile-based installations, including package managers) by installing the actual googler script to libexec, and instead putting a shell wrapper in bin setting
`PYTHONIOENCODING` to `utf-8`.

From now on users can run

```sh
LC_ALL=en_US.US-ASCII PYTHONIOENCODING=ascii googler
```

and expect googler to work as usual.

Single file installations are still vulnerable, but at least 36526b5 makes googler fail noisily with a clear message.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 PREFIX ?= /usr/local
 BINDIR = $(DESTDIR)$(PREFIX)/bin
+LIBEXECDIR = $(DESTDIR)$(PREFIX)/libexec/googler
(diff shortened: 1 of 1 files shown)


jarun:
I think we can skip this one for a while (given that we have a nice message indicating the problem already). Let's wait and see if users are coming back with the same problem. I would keep redirection as the last option.

zmwangx:
Sure.

PR #155, opened 2016-12-17
Improve autocorrect handling: detect autocorrected query and add a prompt command "exact"

### LINKED ISSUE #154: Discussion: remove the --exact option and make it default
Say you run `googler googler`. The word "googler" doesn't even appear once in the results. What the ...? Oh, Google is showing results for "google". But there's no indication whatsoever of this autocorrect in our output; sure, we can introduce more complexity in the parser to detect "Showing results for ...", but just like the result count, this doesn't sit well with our entry-based parser and printer, and is especially problematic with serialized output (`--json`), unless we completely break the interface.

Moreover, autocorrection simply isn't what people expect from command line utilities (autocorrect in interactive shell's line editor is a different story; we're talking about command lines being `exec`'ed here). When you `cat google`, `cat` doesn't autocorrect that to `googler`. You're responsible for what you type. If you made a typo, no sweat, do it again. (You can even correct the query right within googler's "shell", with all options preserved.)

Therefore, I propose to remove the `--exact` option and make it the default (that is, the only) behavior.

@jarun I'll raise a PR if you support this idea.

zmwangx:
Details are in commit messages and additions to parser docs. This PR should be considered on a commit-by-commit basis, and each of the latter two commits (cdfe3d0 and ac6e13c) is optional, although I think they are worthy improvements.

Demo: https://asciinema.org/a/2veowecoo2hpznua9k1bgsnxj.

Closes #154.

### DIFF
--- a/googler
+++ b/googler
@@ -1792,12 +1792,23 @@ class GooglerCmd(object):
 
         """
         self.fetch()
+        colors = self.colors
         if self._autocorrected_to:
-            printerr('\n** Showing results for %s; enter "exact" for an exact search.' %
-                     self._autocorrected_to)
+            autocorrect_info = ('Showing results for %s; enter "exact" for an exact search.' %
+                                self._autocorrected_to)
+            printerr('')
+            if colors:
+                printerr(colors.prompt + autocorrect_info + colors.reset)
+            else:
+                printerr('** ' + autocorrect_info)
         self.display_results(prelude=prelude, json_output=json_output)
         if self._results_filtered:
-            printerr('** Enter "unfilter" to show similar results Google omitted.')
+            unfilter_info = 'Enter "unfilter" to show similar results Google omitted.'
+            if colors:
+                printerr(colors.prompt + unfilter_info + colors.reset)
+            else:
+                printerr('** ' + unfilter_info)
+            printerr('')
 
     def read_next_command(self):
         """Show omniprompt and read user command line.
--- a/googler
+++ b/googler
@@ -1038,6 +1038,31 @@ class GoogleParser(html.parser.HTMLParser):
     #     abstract text
     #   </div>            <!-- stop_populating_textbuf, pop to sitelink abstract -->
     #
+    # - Sometimes Google autocorrects a query. Whenever this happens
+    #   there will be a block whose English version reads "Showing
+    #   results for ... <newline> Search instead for ...", and the HTML
+    #   looks like
+    #
+    #   <span class="spell">Showing results for</span>
+    #   <a class="spell" href="/search?q=google..."><b><i>google</i></b></a>
+    #   <br>
+    #   <span class="spell_orig"></span>
+    #
+    #   We collect the text inside a.spell as the suggested spelling
+    #   (self.suggested_spelling).
+    #
+    #   Note that:
+    #
+    #   1. When npfr=1 (exact), there could still be an
+    #      a.spell, in a block that reads (English version) "Did you mean:
+    #      ...". Therefore, we only consider the query autocorrected when a
+    #      meaningful .spell_orig is also present (self.autocorrected).
+    #
+    #   2. A few garbage display:none, empty tags related to spell
+    #      appear to be always present: span#srfm.spell, a#srfl.spell,
+    #      span#sifm.spell_orig, a#sifl.spell_orig. We need to exclude
+    #      the ids srfm, srfl, sifm and sifl from our consideration.
+    #
     # - Sometimes Google omits similar (more like duplicate) result
     #   entries. Whenever this happens there will be a notice in p#ofr. The way
     #   to unfilter is to simply add '&filter=0' to the query string.
@@ -1125,12 +1150,18 @@ class GoogleParser(html.parser.HTMLParser):
     #           </span>
     #           <br><b>3 HOURS Best Relaxing Music</b> &#39;Romantic <b>Piano</b>&quot; Background <b>Music</b> for Stress ... 3:03 <b>...</b>
     #       </span>
+    #
+    #   8. There's no a.spell_orig when the query is autocorrected; the
+    #      <a> tag (linking to the exact search) is wrapped in the
+    #      span.spell_orig.
 
     def __init__(self, news=False):
         html.parser.HTMLParser.__init__(self)
 
         self.news = news
 
+        self.autocorrected = False
+        self.suggested_spelling = None
         self.filtered = False
         self.results = []
 
@@ -1175,13 +1206,23 @@ class GoogleParser(html.parser.HTMLParser):
             self.set_handlers_to('result')
             return 'result'
 
+        # Autocorrect
+        if tag == 'span' and 'spell_orig' in self.classes(attrs) and attrs.get('id') != 'sifm':
+            self.autocorrected = True
+            return
+        if tag == 'a' and 'spell' in self.classes(attrs) and attrs.get('id') != 'srfl':
+            self.start_populating_textbuf()
+            return 'spell'
+
         # Omitted results
         if tag == 'p' and attrs.get('id') == 'ofr':
             self.filtered = True
 
     @retrieve_tag_annotation
     def root_end(self, tag, annotation):
-        pass
+        if annotation == 'spell':
+            self.stop_populating_textbuf()
+            self.suggested_spelling = self.pop_textbuf()
 
     @annotate_tag
     def result_start(self, tag, attrs):
@@ -1649,6 +1690,7 @@ class GooglerCmd(object):
         atexit.register(self._conn.close)
 
         self.results = []
+        self._autocorrected_to = None
         self._results_filtered = False
         self._urltable = {}
 
@@ -1693,6 +1735,7 @@ class GooglerCmd(object):
         parser.feed(page)
 
         self.results = parser.results
+        self._autocorrected_to = parser.suggested_spelling if parser.autocorrected else None
         self._results_filtered = parser.filtered
         self._urltable = {}
         for r in self.results:
@@ -1749,6 +1792,9 @@ class GooglerCmd(object):
 
         """
         self.fetch()
+        if self._autocorrected_to:
+            printerr('\n** Showing results for %s; enter "exact" for an exact search.' %
+                     self._autocorrected_to)
         self.display_results(prelude=prelude, json_output=json_output)
         if self._results_filtered:
             printerr('** Enter "unfilter" to show similar results Google omitted.')
@@ -1834,9 +1880,16 @@ class GooglerCmd(object):
 
     @require_keywords
     @no_argument
+    def do_exact(self):
+        # Reset start to 0 when exact is applied.
+        self._google_url.update(start=0, exact=True)
+        self.fetch_and_display()
+
+    @require_keywords
+    @no_argument
     def do_unfilter(self):
         # Reset start to 0 when unfilter is applied.
-        self._google_url.update(None, **{'start':0,})
+        self._google_url.update(start=0)
         self._google_url.set_queries(filter=0)
         self.fetch_and_display()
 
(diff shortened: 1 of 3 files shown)


zmwangx:
This "exact" command is like a poor man's version of the originally planned dynamic option manipulation. Just saying.

googler
```diff
@@ -1175,13 +1206,22 @@ class GoogleParser(html.parser.HTMLParser):
             self.set_handlers_to('result')
             return 'result'
 
+        # Autocorrect
+        if 'spell_orig' in self.classes(attrs) and attrs.get('id') not in {'sifm', 'sifl'}:
+            self.autocorrected = True
+        if tag == 'a' and 'spell' in self.classes(attrs) and attrs.get('id') != 'srfl':
```
jarun: Can we make add it as an `elif` case to line 1191: `if tag == 'div' and` ...? A tag can't be both `div` and `a` at the same time.

googler
```diff
@@ -1175,13 +1206,22 @@ class GoogleParser(html.parser.HTMLParser):
             self.set_handlers_to('result')
             return 'result'
 
+        # Autocorrect
+        if 'spell_orig' in self.classes(attrs) and attrs.get('id') not in {'sifm', 'sifl'}:
```
jarun: Can we put this inside a tag check somewhere to avoid `'spell_orig' in self.classes(attrs)` every time?

jarun:
In fact, as I see now, we have ample scope to optimize our parser if we use `elif`. There are many functions where we use if ... if ... if checks. Why not make them elif if they are mutually exclusive? For example: `sitelink_start()`, `title_start()`, `result_end()`, `result_start()` etc.

googler
zmwangx: I'll throw a span tag here. In the beginning it was an a tag, just like `a.spell`, but that doesn't work with `--noua`.

zmwangx:
Addressed comments.

> we have ample scope to optimize our parser if we use elif

As mentioned in an inline comment, because we always return from if bodies, using elif hardly ever helps. Also, the conditions are seldom mutually exclusive, at a glance.

googler
```diff
@@ -1175,13 +1206,23 @@ class GoogleParser(html.parser.HTMLParser):
             self.set_handlers_to('result')
             return 'result'
 
+        # Autocorrect
+        if tag == 'span' and 'spell_orig' in self.classes(attrs) and attrs.get('id') != 'sifm':
+            self.autocorrected = True
+            return
+        if tag == 'a' and 'spell' in self.classes(attrs) and attrs.get('id') != 'srfl':
```
jarun: Doesn't this condition alone indicate auto-correction? I am asking because we already know `spell_orig`... we are the ones passing the search keywords. If possible, I'ld love to skip searching for it.

Also, the reason I am behaving like a miser here is: we do around 7000 pattern matches and 800+ tag parsing for a page on an average.

jarun:
Yes, we are returning in most of the `if`s. I missed this earlier.

googler
zmwangx: > Doesn't this condition alone indicate auto-correction?

It doesn't. See parser docs:

```
    #   1. When npfr=1 (exact), there could still be an
    #      a.spell, in a block that reads (English version) "Did you mean:
    #      ...". Therefore, we only consider the query autocorrected when a
    #      meaningful .spell_orig is also present (self.autocorrected).
```

> we do around 7000 pattern matches and 800+ tag parsing for a page on an average.

I recall a benchmark earlier demonstrating that I/O is the real bottleneck.

jarun: It's not a `bottleneck`, just trying to keep it low if we can.

jarun:
I checked the changes in regular as well as `--noua` mode. Both show same results with auto-correction warning. What's not working with `--noua` (you mentioned something)?

Also, how about highlighting the corrected spelling? At least within quotes?

jarun:
And exact is super-useful now! :+1:

googler
zmwangx: > just trying to keep it low if we can.

I'm afraid we can't optimize here. There's a bit of nickle and diming you can do in the conditions at the cost of overall consistency in the code base. Not much you can do otherwise in a linear search for a bunch of things in rather hostile, poorly structured input.

To truly improve the sensitivity to one added condition, what you can do is scan the tree once, cache the results in sort of search tree(s), then select with logarithmic (or better) complexity. It would be super hardcore though, and whether it will actually save time is questionable for what we do. I checked bs4 source code and it's simply doing a linear traversal for finding and selector queries; see [here](https://git.launchpad.net/~vivekchandola/beautifulsoup/+git/trunk/tree/library/bs4/element.py#n1296) and [here](https://git.launchpad.net/~vivekchandola/beautifulsoup/+git/trunk/tree/library/bs4/element.py#n1518). I don't know how JavaScript folks implement CSS selector queries (`querySelectorAll` and simple special cases).

Anyway, this consideration is not very helpful, since processing time is dwarfed by I/O anyway.

zmwangx:
> I checked the changes in regular as well as --noua mode. Both show same results with auto-correction warning. What's not working with --noua (you mentioned something)?

The code as I initially wrote it (following my understanding of the structure of that block, as documented in the new parser comments) didn't work under --noua, because the block's structure is different under --noua. See the added note under "User Agent disabled (differences)".

I have accounted for the difference, so the code works in whatever environment (hopefully).

> Also, how about highlighting the corrected spelling? At least within quotes?

I don't think we [have a suitable color for it](https://github.com/jarun/googler#colors) so coloring isn't an option. I thought about quotes but rejected that idea, because

```
$ googler -n 3 \"googler news\"

Showing results for "google news"; enter "exact" for an exact search.

1 Google News
https://news.google.com/
Comprehensive up-to-date news coverage, aggregated from sources all over the world by Google News.

2 Google News & Weather on the App Store - iTunes - Apple
https://itunes.apple.com/us/app/google-news-weather/id913753848?mt=8
Oct 3, 2016 - Your comprehensive and personalized view of headline stories and local news & weather. ° Coverage from
75,000 publications ◦ Instant-load ...

3 Google News - Mashable
http://mashable.com/category/google-news/
Google is killing Reader tomorrow, and as a sort of consolation prize, they've added some new features to Google
News. There's now a box that features a ...

googler (? for help) q
```

zmwangx:
Just realized I pasted the wrong demo link...

https://asciinema.org/a/2veowecoo2hpznua9k1bgsnxj

The query above (demonstrating the quoting dilemma) is shown there.

zmwangx:
Actually, we still have one color-neutral styling trick up our sleeves: the underline. Patch on top:

```diff
From 752af338ccfcc5af6d3822c658c7e65d94f4a6f9 Mon Sep 17 00:00:00 2001
From: Zhiming Wang <zmwangx@gmail.com>
Date: Sat, 17 Dec 2016 02:48:58 -0500
Subject: [PATCH] googler: underline autocorrected keyword(s)

---
 googler | 8 +++++++-
 1 file changed, 7 insertions(+), 1 deletion(-)

diff --git a/googler b/googler
index 1768017..2fc3e68 100755
--- a/googler
+++ b/googler
@@ -1794,8 +1794,14 @@ class GooglerCmd(object):
         self.fetch()
         colors = self.colors
         if self._autocorrected_to:
+            if colors:
+                # Underline the keywords
+                print('hi')
+                autocorrected_to = '\x1b[4m' + self._autocorrected_to + '\x1b[24m'
+            else:
+                autocorrected_to = self._autocorrected_to
             autocorrect_info = ('Showing results for %s; enter "exact" for an exact search.' %
-                                self._autocorrected_to)
+                                autocorrected_to)
             printerr('')
             if colors:
                 printerr(colors.prompt + autocorrect_info + colors.reset)
-- 
2.11.0

```

I can apply this if you think it's better.

jarun:
Underline... exactly my thought. I scrolled down to comment and found you've already added it. I'm good with underline. Please update the commit.

zmwangx:
Pushed.

jarun:
Noticed something while testing if this works for new keywords from omniprompt as well. That works but in some cases the correction of spelling is not indicated.

`googler nevy`

show results for navy. But the auto-correction info is not shown in our output. However, I can see in web search that it says: "Including results for navy
Search only for nevy"

This is **inclusion**, not correction.

jarun:
Do you think it's reasonable to handle all these cases or leave it to the user to correct in the omniprompt... now that we have cmdline arg correction in place?

jarun:
Plus the new `exact` keyword to search the last keywords as it is?

zmwangx:
> "Including results for navy Search only for nevy"

When I search for nevy I only get results for nevy... Need the HTML from debug. I'll take a look tomorrow.

> Do you think it's reasonable to handle all these cases or leave it to the user to correct in the omniprompt... now that we have cmdline arg correction in place?

Not sure without digging in, but I dont' think including extra results hurt, and not showing auto-correction info in this case is at least justifiable.

jarun:
OK. I'll dump the results. I think it's also possible that the detection overhead is less now that we have something in place. It falls in `spell` for example.

Here you go: https://pastebin.ubuntu.com/23641696/

zmwangx:
Can't download as text. Put in a gist?

jarun:
Search for "including results for"... I'm also leaving now. We'll conclude this tonight.

zmwangx:
Major changes are required to get this to work because the a tags don't have classes.

1. Need to start from span, and register a custom set of handlers;
2. Need to distinguish between the two cases and pass around more states;
3. More complexity at printing stage.

I don't feel like handling this case. More results doesn't hurt (and this should only happen when the the quality of the results for the original keywords isn't good), unlike results about something else.

jarun:
> Major changes are required to get this to work because the a tags don't have classes.

We drop it then. I'm pulling in the current PR. :)

jarun:
Also I' assuming that we didn't add this to omniprompt help because we show the message explicitly.

zmwangx:
> Also I' assuming that we didn't add this to omniprompt help because we show the message explicitly.

Yes, just like unfilter. The longer the help the less likely it is going to be read.

jarun:
:+1:

jarun:
Can you think of any simple way to navigate using the left/right arrow keys? Or any other key (doesn't have to be noecho keys) combination...

jarun:
It has to be noecho keys... otherwise, if we set n and p we can't search a keyword starting using n and p.

zmwangx:
You can figure out the raw input sequences arrow keys represent (if you ditch readline, otherwise they'll just be captured), but using arrow keys (or ^P, ^N, etc.) for control doesn't sound right outside curses.

zmwangx:
> if we set n and p we can't search a keyword starting using n and p.

Not sure what that means. You can certainly search keywords starting with n or p. The only prohibited keywords are the letters n and p and other commands, verbatim, which are not very useful keywords to begin with.

jarun:
I am trying to skip pressing Enter. Press LEFT ARROW, it goes to prev page, RIGHT ARROW goes to next.

jarun:
Yes, we can't drop readline. Makes life much easier.

jarun:
Also, instead of `enter "exact" for an exact search.`, how about `enter "x" for an exact search.`? Keep it short and sweet.

zmwangx:
> Also, instead of enter "exact" for an exact search., how about enter "x" for an exact search.? Keeps it short and sweet.

I was modeling the "exact" command on "unfilter", but no objection to "x".

jarun:
Please push directly.

jarun:
Also, how about **disabling** the deb package generation now that we are on Ubuntu and have a well-maintained PPA too?

zmwangx:
> Also, how about disabling the deb package generation now that we are on Ubuntu and have a well-maintained PPA too?

I'm writing the commit message right now :)

zmwangx:
> Please push directly.

4827b5de0ead05a8cf5a309ddd9f0d6f91e19dff.

PR #156, opened 2016-12-17
Miscellaneous improvements

zmwangx:
- Add -V, --version option (standard stuff);
- Reorder options in completion defs according to help text, so that checking whether any option's missing is easier;
- Add Debian and Ubuntu availability badges (can't give a version, so just point out which distro releases googler is available on, and link to respective package search pages).

We may want to cut a release some time soon.


jarun:
> We may want to cut a release some time soon.

How about late today? ;)

zmwangx:
> How about late today? ;)

No problem.

jarun:
:+1:

jarun:
Thank you!

PR #157, opened 2016-12-18
Remove deb package building and mention

zmwangx:
googler has been included in Debian Stretch and Ubuntu Yakkety. There's also a PPA listed. Most people should just use the official repos or the PPA.

Those who want to use an independent, latest version should choose another installation method, e.g., download the single script.


tools/makedeb
```diff
@@ -1,277 +0,0 @@
-#!/usr/bin/env zsh
```
zmwangx: It's been fun. In a sense.

jarun:
Thank you!

PR #161, opened 2016-12-28
Make it possible to install Googler with pip (python setup.py install)

jorgesumle:
The only thing left is to review the code and make it able to install the man pages and the auto-completion stuff automatically with pip. One Python project that does that is YouTube-dl. Can anyone guide me or help me with that?


jarun:
Hi,

The devs have taken a conscious decision not to submit `googler` to PyPi earlier.

jorgesumle:
OK, I respect that.

PR #171, opened 2017-03-02
Include commit hash in version number when possible

zmwangx:
- In a git repo (that is, when `googler` — after symlink resolution — is a git repo), use `git-describe` (or `git-rev-parse` as fallback) on HEAD to determine the exact commit. Examples:

  ```
  $ googler --version
  2.9-21-g3f46a90
  ```

  or when the clone is shallow, without tags:

  ```
  $ ./googler --version
  2.9-*-g3f46a90
  ```

  Note that if HEAD is at an exact tag, you won't see that trailing `-<count>-g<sha>` (which is apparently preferable), simply due to how `git describe --abbrev HEAD` works.

- Outside a git repo, if the script is downloaded via `googler --upgrade --include-git`, we now put the commit hash into `_VERSION_`, also in the `<major>.<minor>-*-g<sha>` format seen in the shallow case above (since we can't get commit count since tag/release from GitHub API v3). Sample session:

  ```
  $ ./googler -u --include-git
  Upgrading from version 2.9-*-g3f46a90
  Downloading https://raw.githubusercontent.com/jarun/googler/df256cf083488be6aa9ea4459b624e84802ab0df/googler
  Upgraded to version 2.9-*-gdf256cf.
  ```

  A nice bonus is we can now warn about downgrading from git master to stable (due to missing `--include-git`:

  ```
  $ ./googler -u
  Upgrading from version 2.9-*-g3f46a90
  Downloading https://raw.githubusercontent.com/jarun/googler/v2.9/googler
  [WARNING] 2.9-*-g3f46a90 => 2.9 may be a downgrade; did you forget --include-git?
  Upgraded to version 2.9.
  ```

  To be fair, maybe zero people will benefit from this (I'm afraid at most a handful users use `-u`, less `--include-git`), but IMO it's pretty neat from a design point of view. 💯

### DIFF
--- a/googler
+++ b/googler
@@ -99,6 +99,36 @@ RAW_DOWNLOAD_REPO_BASE = 'https://raw.githubusercontent.com/jarun/googler'
 
 # Global helper functions
 
+def _version_():
+    """Returns a version number with git commit embedded if possible.
+
+    Falls back to _VERSION_ otherwise.
+    """
+    import subprocess
+
+    script_dir = os.path.dirname(os.path.realpath(__file__))
+    DEVNULL = subprocess.DEVNULL
+
+    try:
+        description = subprocess.check_output(
+            ['git', '-C', script_dir, 'describe', '--abbrev', 'HEAD'],
+            stdin=DEVNULL, stderr=DEVNULL,
+        ).decode('utf-8')
+        return description.lstrip('v')
+    except subprocess.CalledProcessError:
+        pass
+
+    try:
+        sha = subprocess.check_output(
+            ['git', '-C', script_dir, 'rev-parse', '--short', 'HEAD'],
+            stdin=DEVNULL, stderr=DEVNULL,
+        ).decode('utf-8')
+        return '%s-*-g%s' % (_VERSION_, sha)
+    except subprocess.CalledProcessError:
+        pass
+
+    return _VERSION_
+
 def open_url(url):
     """Open an URL in the user's default web browser.
 
@@ -2012,7 +2042,7 @@ class GooglerArgumentParser(argparse.ArgumentParser):
         Zhiming Wang <zmwangx@gmail.com>
         License: GPLv3
         Webpage: https://github.com/jarun/googler
-        """ % _VERSION_))
+        """ % _version_()))
 
     # Augment print_help to print more than synopsis and options
     def print_help(self, file=None):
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

jarun:
It shows the version only within the source directory. Will it be really useful at any point in time?

> A nice bonus is we can now warn about downgrading from git master to stable

People would prefer using standalone. This is what I am seeing:

```
~/G/googler$ ./googler -v
2.9-19-ge3ebac8
~/G/googler$ cp googler ~/
~/G/googler$ cd
~$ ./googler -v
2.9
~$ ./googler -u
Upgrading from version 2.9
Downloading https://raw.githubusercontent.com/jarun/googler/v2.9/googler
Upgraded to version 2.9.
```

So the same file returns to v 2.9 outside the source directory and downgrades to v2.9 without any check. I think this would be more confusing than useful if at all someone is using `googler` from the source directory. If he ever installs, he gets v2.9 for the same file! Initially I thought it's anice addition, but now I see more complexity in understanding the behaviour. We should keep it simple.

zmwangx:
> It shows the version only within the source directory. Will it be really useful at any point in time?

Useful for me at least. Or any one which symlinks to the source directory.

> So the same file returns to v 2.9 outside the source directory and downgrades to v2.9 without any check.

Because it's impossible to check. That was the case too before this PR, so it's not a regression.

You only get the benefit it you stick to one installation method. That is, you either download stable and upgrade to stable, or download git master and upgrade to master. Downloading master first then trying to "upgrade" to stable doesn't help, and it's been the case all along.

zmwangx:
> You only get the benefit it you stick to one installation method. That is, you either download stable and upgrade to stable, or download git master and upgrade to master.

Actually that's not entirely accurate. Downloading stable then upgrading to git master is perfectly fine. But not the other way round.

zmwangx:
By the way, a little bit of context for improving version detection: it's really *really* hard to get people to provide adequate debugging information, even when you explicitly hand over an itemized list of things you need (proven time and again). So it's definitely a win if you can extract more of that info from a single command, rather than ask separately.

jarun:
> a little bit of context for improving version detection

At least I don't remember any issue where we had trouble because we didn't know the commit hash for `googler`. I understand people do not provide info we need many a time, but not knowing commit hash was hardly an issue ever. So my take here is we are overdoing it.

We can't tell users to install git or provide us the file he is using as a general rule. Probably 99% of the time they will tell us the release version.

zmwangx:
> So my take here is we are overdoing it.

Sure, that's a valid argument. If you're bothered by the complexity, feel free to roll back. (But do keep 6a3c025, that's an independent fix.)

> Probably 99% of the time they will tell us the release version.

Possible.

jarun:
Actually I think we should add a `risky` in our README where we say how to upgrade to master to show that we prefer regular users to use the released version.

zmwangx:
👍

PR #173, opened 2017-03-13
Display results after returning from browser

demonye:
Current search results disappeared after quitting from w3m, and I didn't remember the index of another page I was interested. I think the current result should be displayed when returning back from the browser.

### DIFF
--- a/googler
+++ b/googler
@@ -1966,8 +1966,10 @@ class GooglerCmd(object):
                     self.do_next('')
                 elif cmd == 'o':
                     self.do_open()
+                    self.display_results()
                 elif cmd.startswith('o '):
                     self.do_open(*cmd[2:].split())
+                    self.display_results()
                 elif cmd == 'p':
                     self.do_previous('')
                 elif cmd == 'q':
@@ -1980,6 +1982,7 @@ class GooglerCmd(object):
                     self.help()
                 elif cmd in self._urltable:
                     open_url(self._urltable[cmd])
+                    self.display_results()
                 elif self.keywords and cmd.isdigit() and int(cmd) < 100:
                     printerr('Index out of bound. To search for the number, use g.')
                 else:


zmwangx:
> Current search results disappeared after quitting from w3m

It's not clear to me what's causing this. We never clear screen for you, and my w3m 0.5.3 (running on either macOS 10.12.3 or Ubuntu 16.04) doesn't clear screen on exit either. Maybe you somehow configured w3m to clear screen on exit?

Anyway, this PR is not mergeable at its current state, because reprinting everything unconditionally every time something is opened is wrong.

demonye:
Well, I found the reason which is because I exported TERM=linux I will see if I have better solution. Thanks for the reply.

jarun:
Thanks for the update!

zmwangx:
I can confirm that `TERM=linux` does lead to this behavior.

Looking at w3m source code, specifically `terms.c` (they don't use curses by the way), here are the relevant bits:

```c
#define GETSTR(v,s) {v = pt; suc = tgetstr(s,&pt); if (!suc) v = ""; else v = allocStr(suc, -1); }

...

void
reset_tty(void)
{
    writestr(T_op);             /* turn off */
    writestr(T_me);
    if (!Do_not_use_ti_te) {
        if (T_te && *T_te)
            writestr(T_te);
        else
            writestr(T_cl);
    }
    writestr(T_se);             /* reset terminal */
    flush_tty();
    TerminalSet(tty, &d_ioval);
    close_tty();
}

...

    GETSTR(T_cr, "cr");         /* carriage return */
    GETSTR(T_ta, "ta");         /* tab */
    GETSTR(T_sc, "sc");         /* save cursor */
    GETSTR(T_rc, "rc");         /* restore cursor */
    GETSTR(T_so, "so");         /* standout mode */
    GETSTR(T_se, "se");         /* standout mode end */
    GETSTR(T_us, "us");         /* underline mode */
    GETSTR(T_ue, "ue");         /* underline mode end */
    GETSTR(T_md, "md");         /* bold mode */
    GETSTR(T_me, "me");         /* bold mode end */
    GETSTR(T_cl, "cl");         /* clear screen */
    GETSTR(T_cm, "cm");         /* cursor move */
    GETSTR(T_al, "al");         /* append line */
    GETSTR(T_sr, "sr");         /* scroll reverse */
    GETSTR(T_ti, "ti");         /* terminal init */
    GETSTR(T_te, "te");         /* terminal end */
    GETSTR(T_nd, "nd");         /* move right one space */
    GETSTR(T_eA, "eA");         /* enable alternative charset */
    GETSTR(T_as, "as");         /* alternative (graphic) charset start */
    GETSTR(T_ae, "ae");         /* alternative (graphic) charset end */
    GETSTR(T_ac, "ac");         /* graphics charset pairs */
    GETSTR(T_op, "op");         /* set default color pair to its original value */

...
```

As you can see, unless you pass the `-X` flag (`Do_not_use_ti_te`), w3m writes a `T_te` (termcap `te`, terminal end) if the terminal is capable of it, or a `T_cl` (termcap `cl`, clear screen) otherwise. In the case of `TERM=linux`, the `te` capability is not available (you can easily check the termcap database), so a `T_cl` is written. See `termcap(5)` for more info: https://linux.die.net/man/5/termcap.

zmwangx:
I'm not sure what's the best course of action here. We're clearly not in the business of detecting termcaps and figuring out how different tty browsers respond to different termcaps; at the same time, OP's pain is very real, I can understand that. Maybe we can add a flag to allow re-listing after a tty browser quits. But before we discuss that option, let me ask this question @demonye:

What's your operating system and terminal emulator? Is your terminal emulator really only capable of `TERM=linux`? If so, have you considered getting yourself a better terminal emulator? I looked into a whole bunch of `TERM`s, including `xterm*`, `gnome`, `rxvt`, `konsole`, `putty`, `screen`, etc., and they all support `ti` and `te`.

demonye:
Hi Zhiming,

Thanks for the kind email, I'm using ubuntu 16.04 and default gnome
terminal. No worry that I can set alias like g='TERM=xterm googler' to
solve my problem. The reason I set TERM to linux is when I exit vim I'd
like the file content still display on the screen.

Redards,
Eric
​

zmwangx:
Cool, so your terminal is capable enough, it's only crippled by your conscious decision of setting TERM to linux, and there's an easy fix. I'm closing this PR then — thanks anyway.

Let's think about taking action only if another user raises the same issue (and is not satisfied with the solution here).

PR #174, opened 2017-03-14
README: set Cache-Control: max-age=600 on various version badges

zmwangx:
Previously they had the following `Cache-Control` header:

    Cache-Control: no-cache, no-store, must-revalidate

which means a refresh — and a visual flicker — on every single page refresh. That's not really necessary.


jarun:
Thank you!

PR #176, opened 2017-03-17
Convention fix

jorgesumle:
Use always single quotes.


jarun:
This is a very small change (and not a bug fix at that) to be merged. We can push it in with one of the bugger changes sometime. Do you have any such change in mind?

jarun:
I have pushed the change with commit 0cbdd623365b6e5923d6144bb6f86809e257f4f4.

PR #178, opened 2017-03-18
Miscellaneous

zmwangx:
By the way, I'm not sure our use of `OR` is well-defined. According to https://support.google.com/websearch/answer/2466433?hl=en:

> Combine searches
> Put "OR" between each search query. For example,  marathon OR race.
>
> Search for a specific site
> Put "site:" in front of a site or domain. For example, site:youtube.com or site:.gov.

It's not entirely clear what consists of a "search query", but in this context I think it means the keywords/keyphrases without qualifiers; at least that's the only safe way to read it. Also, it's not clear which parts of a query are or'ed (that is, where to put parentheses if you were to parse the query), in a query like

    keyword filetype:type site:example.com OR site:example.org OR site:example.net

https://www.google.com/advanced_search also does not provide a way to search multiple sites/domains.

Therefore, my conclusion is we're treading on a less well-defined rule, but it appears to work:

```
$ googler --debug --json -w github.com -w gitlab.com iterm2 | jq -r '.[].url'
[DEBUG] Version 3.0
[DEBUG] Connecting to new host www.google.com
[DEBUG] Fetching URL /search?ie=UTF-8&nfpr=1&num=10&oe=UTF-8&q=iterm2+site:github.com+OR+site:gitlab.com
[DEBUG] Cookie: NID=99=QU3W5DneIxaOl5nxPadSh_GEaBMIZ9DE0-oX6wV2FWAVoih_E5czpZBz992il4BzGzmQ4FcVfoUFwF9xIJsxtZ1miWdNuLimlN7srcaPDQvCPolJjLQX8Xo27LU_ua50jCRPUEjBUxWxKKr6
[DEBUG] Response body written to '/Volumes/ramdisk/googler-response-6_ds7_kl'.
https://github.com/gnachman/iTerm2
https://github.com/altercation/solarized/tree/master/iterm2-colors-solarized
https://github.com/mbadolato/iTerm2-Color-Schemes
https://gist.github.com/kevin-smets/8568070
https://github.com/chriskempson/base16-iterm2
https://github.com/morhetz/gruvbox-contrib/tree/master/iterm2
https://github.com/mhartington/oceanic-next-iterm
https://github.com/mmastrac/iterm2-zmodem
https://gitlab.com/gnachman/iterm2
https://gitlab.com/gnachman/iterm2/issues
```

I was thinking about making it clear in docs that `-w, --site` can be specified multiple times, but given the arguments above, it's probably safer to keep it an undocumented feature. After all, I'm not prepared to spend even more time on a feature that one user has requested over the lifetime of googler.

### DIFF
--- a/googler
+++ b/googler
@@ -292,7 +292,7 @@ class GoogleUrl(object):
         self._num = 10
         self._start = 0
         self._keywords = []
-        self._site = None
+        self._sites = None
         self._query_dict = {
             'ie': 'UTF-8',
             'oe': 'UTF-8',
@@ -418,8 +418,8 @@ class GoogleUrl(object):
                 qd.pop('tbm', None)
         if 'num' in opts:
             self._num = opts['num']
-        if 'site' in opts:
-            self._site = opts['site']
+        if 'sites' in opts:
+            self._sites = opts['sites']
         if 'start' in opts:
             self._start = opts['start']
         if 'tld' in opts:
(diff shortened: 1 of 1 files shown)


jarun:
I went by how it appears to work in practicality. My web search string was:

`reboot site:superuser.com OR site:android.stackexchange.com`

and I got:

`https://www.google.co.in/?gfe_rd=cr&ei=TsnMWNLANOPx8AfL16TYBQ&gws_rd=ssl#q=reboot+site:superuser.com+OR+site:android.stackexchange.com&*`

So it appears OR when added explicitly does combine multiple *search refinements*.

jarun:
Thank you!

zmwangx:
> I went by how it appears to work in practicality.

Yep, as I said it appears to work for me too. Just don't want to explicitly mention this in the docs. (I had a commit to add "can be used multiple times" to the docs which I scrapped.)

Issue #179, opened 2017-03-25
running on bash on windows 10 gives error


ali4ever4:
Hello i just tried this on bash on windows 10 and got the following error when try to do anything with googler

[ERROR] Failed to connect to www.google.com: [Errno 22] Invalid argument

can someone help ?

zmwangx:
You've been given an issue template: https://github.com/jarun/googler/blob/master/.github/ISSUE_TEMPLATE.md. Please don't ignore it.

ali4ever4:
ok thanks 

- Output of googler -d;

[DEBUG] Version 3.0
[DEBUG] Connecting to new host www.google.com
Traceback (most recent call last):
  File "/usr/bin/googler", line 682, in new_connection
    self._conn.connect(self._notweak)
  File "/usr/bin/googler", line 213, in connect
    sock.setsockopt(socket.SOL_TCP, socket.TCP_DEFER_ACCEPT, 1)
OSError: [Errno 22] Invalid argument

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/googler", line 2357, in <module>
    main()
  File "/usr/bin/googler", line 2331, in main
    repl = GooglerCmd(opts)
  File "/usr/bin/googler", line 1703, in __init__
    notweak=opts.notweak)
  File "/usr/bin/googler", line 633, in __init__
    self.new_connection(host, port=port, timeout=timeout)
  File "/usr/bin/googler", line 685, in new_connection
    raise GoogleConnectionError(msg)
__main__.GoogleConnectionError: Failed to connect to www.google.com: [Errno 22] Invalid argument.

- Details of operating system
Windows 10 , bash , python version 2.7

zmwangx:
That's guarded by a `sys.platform.startswith('linux')` conditional. I guess you're using Windows' Linux subsystem? Apparently they're not all the way there yet. I guess we can ignore errors from `setsockopt`.

> python version 2.7

googler doesn't work with Python 2.7. README:

> googler requires Python 3.3 or later.

zmwangx:
#180. Please test if you wish.

jarun:
#180 is merged now. Please test and confirm.

ali4ever4:
thanks it works perfect now

ali4ever4:
now need to find a way for it to understand how to open windows browser from linux subsystem bash -_- 

i guess that is still not possible in the current stable update of win 10 , not sure

zmwangx:
[...]

PR #180, opened 2017-03-25
Proceed when socket.setsockopt fails

zmwangx:
According to user report (#179), Windows 10's Linux subsystem claims to be Linux but doesn't seem to support the socket options here.


jarun:
Thank you!

PR #182, opened 2017-04-02
Correction in Bash completion file

jorgesumle:
The variable declared with `local -a` was not being used.


jarun:
As this is a very minor change we can push it in the next time we fix something in `googler`.

jarun:
or even ignore...

PR #190, opened 2017-06-23
Handle basic HTTP auth when using proxy

bobey:
Fixes #189

PR #191, opened 2017-06-25
More proxy stuff: clean up after #190 and improve upon that

zmwangx:
- Fix the mistakes of #190:
  - Logic was screwed up: what's `if proxy.find('@')`? -1 is truthy.
  - User and password not percent-decoded.

- Bring proxy spec up to par with curl: `[http://][user:password@]proxyhost[:port]`;

- Update docs. Notably, we were talking about HTTPS proxies while we never supported HTTPS connections to proxies (my fault I believe — I was confusing various terms when I first implemented proxy support); that has been fixed — now we only talk about HTTP proxies, and an explicit `https://` scheme even results in a `NotImplementedError`.

### DIFF
--- a/googler
+++ b/googler
@@ -672,22 +672,18 @@ class GoogleConnection(object):
         proxy = self._proxy
 
         if proxy:
-            auth_needed = False
-            if proxy.find('@'):
-                auth_needed = True
-                proxy_host_port = proxy[proxy.find('@')+1:]
-                proxy_user_pwd = proxy[0:proxy.find('@')]
+            proxy_user_passwd, proxy_host_port = parse_proxy_spec(proxy)
 
-            logger.debug('Connecting to proxy server %s', proxy)
+            logger.debug('Connecting to proxy server %s', proxy_host_port)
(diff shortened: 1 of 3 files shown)


googler
```diff
@@ -2253,16 +2249,35 @@ def self_upgrade(include_git=False):
...
-        return None
-    proxy = urllib.parse.urlparse(os.environ['https_proxy']).netloc
-    # urlparse recognizes a netloc only if it is introduced by '//'
-    # (https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse).
-    # This means a bare host:port like 'localhost:8118' won't be
-    # recognized. In that case, try again with the '//' prefix.
-    if not proxy:
-        proxy = urllib.parse.urlparse('//' + os.environ['https_proxy']).netloc
-    return proxy if proxy else None
+    return os.getenv('https_proxy')
```
zmwangx: A nice side effect is that the `https_proxy` env var can be used directly without separate parsing.

jarun:
Thank you!

Issue #193, opened 2017-07-05
unfilter results without typing "unfilter"


mattmar:
Hi all,
I need to perform a full Google search from the command line and  therefore I am using Googler.
However when I run:
`googler -x -C -n 500 -s 100 -c cl -l sp Clínica+Veterinaria+Putre`

I get a few results and then:
> ** Enter "unfilter" to show similar results Google omitted.

Browsing Googler source code I spotted these comments:

>     #   Sometimes Google omits similar (more like duplicate) result
>     #   entries. Whenever this happens there will be a notice in p#ofr. The way
>     #   to unfilter is to simply add '&filter=0' to the query string.

When I add "&filter=0" two the query, I obtain 0 results.
`googler -x -C -n 500 -s 100 -c cl -l sp "Clínica+Veterinaria+Putre&filter=0"`

Can you help me to overcome this problem?
Thanks 
Matteo

jarun:
You shouldn't try to manipulate the search string. There was a PR on this sometime back. Take a look at https://github.com/jarun/googler/pull/188. If you really want it, fork and raise the PR.

jarun:
Closing this issue,

PR #188, opened 2017-06-13
Add --unfilter flag, make implicit on --noprompt

michaelhood:

PR #194, opened 2017-07-06
Add --unfilter option

zmwangx:
Since two people requested this (#188, #193), let's just add this to prevent further requests.


zmwangx:
Also, now that we're not committing regularly, what about a release to flush out the new features?

jarun:
> ... let's just add this to prevent further requests.

Thanks! I'd been meaning to do this yesterday evening. Returned home late and forgot all about it ;).

> Also, now that we're not committing regularly, what about a release to flush out the new features?

Sure thing! I'll release the new version right away.

jarun:
Thanks!

PR #196, opened 2017-07-11
.travis.yml: add py36 target

zmwangx:
Just realized we haven't added the py36 target (IIRC there were some problems in the beta/rc stage).


jarun:
Thank you!

PR #198, opened 2017-08-12
Autocomplete queries

zmwangx:
Using the https://www.google.com/complete/search API. See `completer_fetch_completions` in `googler` for details.

Functionality added to completion scripts for all three shells. No opt-out at the moment.


jarun:
Thank you!

PR #204, opened 2017-09-25
fix prompt color issues - parse colors with print() before using input()

urielz:
works for me. perhaps there's a more elegant way to do this...

### DIFF
--- a/googler
+++ b/googler
@@ -1886,7 +1886,8 @@ class GooglerCmd(object):
         enter_count = 0
         while True:
             try:
-                cmd = input(prompt)
+                sys.stderr.write(prompt)
+                cmd = input()
             except EOFError:
                 sys.exit(0)
 


googler
```diff
@@ -1886,7 +1886,8 @@ class GooglerCmd(object):
         enter_count = 0
         while True:
             try:
-                cmd = input(prompt)
+                print(message,end='')
```
zmwangx: Space after comma.

zmwangx:
In fact, the prompt message should be printed to stderr, so you want to use `printerr` instead. You can add an `end` kwarg to the current `printerr`.

urielz:
@zmwangx thanks. my knowledge of python is close to null. would latest commit work?

zmwangx:
I fixed up and squashed the commits (you're still the author), but I want to make sure I take a look at and understand the bpo issue before letting this in.

By the way, are you using libedit or libreadline? You just need to provide the output of

```sh
python3 -c 'import readline; print(readline.__doc__)'
```

jarun:
@zmwangx please approve and merge when you guys are done. I'm done for the day.

zmwangx:
@jarun Sure, good night.

urielz:
@zmwangx sure, no rush at all with merging this.

Looks like I have it:
python3 -c 'import readline; print(readline.__doc__)'
Importing this module enables command line editing using libedit readline.

zmwangx:
Okay, your readline capabilities are provided by libedit, not libreadline. That might be the difference. I'll compile python without readline tonight to see if I can reproduce.

zmwangx:
I read bpo 17337 and it's totally unrelated. I also compiled current master of python/cpython without readline, and still can't reproduce. So there's that.

Anyway, I'm gonna merge this.

jarun:
Thanks @zmwangx!
Thanks @urielz for your time!

urielz:
thank you guys. googler is pretty neat :)

jarun:
There are 2 issues I found with this change.

1. Start with a fullscreen terminal. Googler something. Then un-maximize the terminal. The prompt is lost. Two screenshots with and without the change below:

promt_gone

2. Search something. Press `\`, then `Bksp`. The prompt will be erased.

These are _serious_ issues wrt. user experience. We should retain the older code. I will revert this change.

zmwangx:
Sure, I don't like this either.

PR #205, opened 2017-09-29
Move tests/devbin contents to tests

zmwangx:
Somewhat loosely speaking, `parse` can be used for tests, if we provide a static set of HTML input and check the output for regressions. We're just not doing that.

Also, the direct cause of this commit is the `googler.py` symlink was broken.


jarun:
Thank you!

PR #208, opened 2017-11-06
Allow passing source address on commandline

kyrias:
We use googler for a bot, and one of the bots IPs occasionally gets blocked by Google.  When that's happened I have previously hardcoded the second IP as the `source_address` argument to `HTTPSConnection`, but being able to pass it on the commandline would make things much simpler.

### DIFF
--- a/googler
+++ b/googler
@@ -666,13 +666,13 @@ class GoogleConnection(object):
 
     """
 
-    def __init__(self, host, port=None, timeout=45, proxy=None, notweak=False):
+    def __init__(self, host, port=None, timeout=45, proxy=None, notweak=False, source_address=None):
         self._host = None
         self._port = None
         self._proxy = proxy
         self._notweak = notweak
         self._conn = None
-        self.new_connection(host, port=port, timeout=timeout)
+        self.new_connection(host, port=port, timeout=timeout, source_address=source_address)
         self.cookie = ''
 
     @property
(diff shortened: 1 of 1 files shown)


jarun:
Any kind of automated scraping is [not allowed by google](https://support.google.com/webmasters/answer/66357?hl=en).

We can't add this to googler.

kyrias:
This doesn't actually automate anything though, it implements an actually useful feature.  There are multiple reasons you might *NEED* to be able to specify the source IP that Python should connect from.  It just *happens* to also be useful for this as well.

(And it's not really any more automated scraping than any other use of this script, it's just remote execution through a bot.)

jarun:
Please help me understand how this new provision helps a regular user at his workstation trying to look up something using googler.

kyrias:
"Regular" I don't know, but I have worked in environments where the default route had to be the secure VPN to reduce the risk of any traffic leaking onto the Internet, and anything that for some reason needed Internet access as well has a secondary IP that is required to be used for connections to Internet resources.

Anyway, we'll just switch to a private fork either way with some of our other enhancements as well.  Have a nice day.

jarun:
> default route had to be the secure VPN

googler supports proxy along with auth.

>  as well has a secondary IP

I have (remotely) worked on this kind of setups myself. However, this is not a very usual _personal_ setup. `googler` has thousands of users but this is the first time I am receiving this request.

I hope you understand the real reason behind my primary reservation. Now that we have a PR, let's hold on for a while and see if some individual users are asking for it. If they do, I'll happily pull it in.

PR #211, opened 2017-11-21
Fix typo in README

jwilk:


jarun:
As this is not immediately visible, I'll make the change during the next commit.

PR #214, opened 2017-12-03
Added Docker feature

jaymoulin:
This feature add docker to ease installation and propose an isolated system. This is completely optional since I propose this on my own (https://github.com/jaymoulin/docker-googler) and images are already compiled.


jarun:
Sorry, `googler` has few dependencies and it's not a library. It doesn't mess with a system either. We can safely skip Docker.

PR #215, opened 2017-12-06
Added support for full name of commands

AidanHanda:
Added so that people can use google instead of g, override instead of O, etc.


jarun:
Why? Doesn't a quick help show you which option means what?

jarun:
I am closing this one. Just not relevant at this stage of the project. If you are interested in contributing please take a list at the first item on ToDo list.

PR #222, opened 2018-03-30
If proxy environment variable has '/' after port number, googler fails.

dilawar:
For proxy environemtn variable such as https_proyx=http://proxy.abc.in:3128/ with leading '/' after port number
, currently googler would not work because '3128/' is not a numeric port. This PR fixes this. This format with '/' works with almost many other tools such as wget, curl etc. '/' after port number is suggested at many places including arch https://wiki.archlinux.org/index.php/proxy_settings .

- ~~~Applicable only for environment variable.~~~
- ~~~If user is passing `--proxy=http://proxy.abc.in:3128/' from command line, it is not applicable.~~~

### DIFF
--- a/googler
+++ b/googler
@@ -2360,7 +2360,7 @@ def parse_proxy_spec(proxyspec):
         pos = proxyspec.find('@')
         user_passwd = urllib.parse.unquote(proxyspec[:pos])
         # Remove trailing '/' if any
-        host_port = proxyspec[pos+1:].replace( r'/', '' ) 
+        host_port = proxyspec[pos+1:].rstrip('/') 
     else:
         user_passwd = None
         host_port = proxyspec
--- a/googler
+++ b/googler
@@ -2340,7 +2340,9 @@ def python_version():
 
 
 def https_proxy_from_environment():
-    return os.getenv('https_proxy')
+    import re
+    # remove leading / from port number if any.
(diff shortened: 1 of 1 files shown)


zmwangx:
Please modify `parse_proxy_spec` instead of `https_proxy_from_environment`. Also, regex's not necessary here; just use `rstrip`.

googler
```diff
@@ -2340,7 +2340,7 @@ def python_version():
 
 
 def https_proxy_from_environment():
-    return os.getenv('https_proxy')
+    return os.getenv( 'https_proxy' )
```
jarun: please remove the redundant spaces.

googler
```diff
@@ -2359,7 +2359,8 @@ def parse_proxy_spec(proxyspec):
     if '@' in proxyspec:
         pos = proxyspec.find('@')
         user_passwd = urllib.parse.unquote(proxyspec[:pos])
-        host_port = proxyspec[pos+1:]
+        # Remove trailing '/' if any
+        host_port = proxyspec[pos+1:].replace( r'/', '' )
```
jarun: please remove the redundant spaces for replace().

jarun: Also, please use rstrip:

    proxyspec[pos+1:].rstrip('/')

googler
jarun: I will take care of this.

jarun: Fixed at 8253de92f8f3c7a38c6069c3e41b559066e75a74.

Issue #230, opened 2018-05-28
add support for tmux buffer


jose1711:
currently googler uses xsel on linux which means it a running Xserver is required to use it. tmux's buffers can be used as a terminal-only solution. it would be great if googler adds support for this (```['tmux', 'setb', self._urltable[cmd[2:]].encode('utf-8')]```)

jarun:
Can you please send a patch? Also, is there a way to determine if we are running on `tmux`?

jarun:
I think the correct check would be to see if `$TERM` starts with `tmux` because there can be multiple strings based on user preference.

Please confirm if the following patch works for you:

```
diff --git a/googler b/googler
index b26dda3..9042af7 100755
--- a/googler
+++ b/googler
@@ -2090,6 +2090,8 @@ class GooglerCmd(object):
                             copier_params = ['pbcopy']
                         elif sys.platform == 'win32':
                             copier_params = ['clip']
+                        elif os.getenv('TERM').startswith('tmux'):
+                            copier_params = ['tmux', 'setb', self._urltable[cmd[2:]].encode('utf-8')]
                         else:
                             copier_params = []
```

zmwangx:
IIRC the correct way to detect tmux is to check the `TMUX_PANE` env var (the `TMUX` env var is probably set, too, but IIRC it's not documented). `TERM` is not a reliable indicator at all; I have

```tmux
set -g default-terminal "screen-256color"
```

in my `.tmux.conf`, for instance.

zmwangx:
By the way, I'm not sure that using a tmux buffer unconditionally within tmux instead of a system-wide clipboard is actually a good idea.

jarun:
Please test the following patch as suggested by @zmwangx:

```
diff --git a/googler b/googler
index b26dda3..1c57ffd 100755
--- a/googler
+++ b/googler
@@ -2090,6 +2090,8 @@ class GooglerCmd(object):
                             copier_params = ['pbcopy']
                         elif sys.platform == 'win32':
                             copier_params = ['clip']
+                        elif os.getenv('TMUX_PANE', None) is not None:
+                            copier_params = ['tmux', 'setb']
                         else:
                             copier_params = []
 

```

zmwangx:
https://github.com/jarun/googler/blob/7b6ad41226d00386a88fe369d02abce43be90a29/googler#L2084-L2094

You'll probably get a `FileNotFoundError` first. Just typing away without testing, but what about:

```py
[...]

PR #231, opened 2018-05-31
Support tmux buffer to copy url (with --tmux)

jose1711:
addresses https://github.com/jarun/googler/issues/230

### DIFF
--- a/auto-completion/zsh/_googler
+++ b/auto-completion/zsh/_googler
@@ -58,6 +58,7 @@ args=(
     '(--include-git)--include-git[when used with --upgrade, upgrade to git master]'
     '(- : *)'{-v,--version}'[show version number and exit]'
     '(-d --debug)'{-d,--debug}'[enable debugging]'
+    '(--tmux)--tmux[use tmux for clipboard operations]'
     '*:::query:_googler_complete_query'
 )
 _arguments -S -s $args
--- a/googler
+++ b/googler
@@ -723,7 +723,7 @@ class GoogleConnection(object):
 
             try:
                 self._conn.connect(self._notweak)
-            except Exception as e:
+            except Exception:
                 msg = 'Failed to connect to proxy server %s: %s.' % (proxy, e)
                 raise GoogleConnectionError(msg)
         else:
@@ -2083,9 +2083,12 @@ class GooglerCmd(object):
                     try:
                         # try copying the url to clipboard using native utilities
                         if sys.platform.startswith(('linux', 'freebsd', 'openbsd')):
+                            if self._opts.tmux and os.getenv('TMUX_PANE'):
+                                copier_params = 'tmux_detected'
+                            else:
+                                copier_params = ['xsel', '-b', '-i']
                             if shutil.which('xsel') is None:
                                 raise FileNotFoundError
-                            copier_params = ['xsel', '-b', '-i']
                         elif sys.platform == 'darwin':
                             copier_params = ['pbcopy']
                         elif sys.platform == 'win32':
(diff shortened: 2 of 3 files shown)


auto-completion/zsh/_googler
```diff
@@ -58,6 +58,7 @@ args=(
     '(--include-git)--include-git[when used with --upgrade, upgrade to git master]'
     '(- : *)'{-v,--version}'[show version number and exit]'
     '(-d --debug)'{-d,--debug}'[enable debugging]'
+    '(--tmux)--tmux[use tmux for clipboard operations]'
```
jarun: No new option for this please. PS my next comment.

googler
```diff
@@ -2083,9 +2083,12 @@ class GooglerCmd(object):
                     try:
                         # try copying the url to clipboard using native utilities
                         if sys.platform.startswith(('linux', 'freebsd', 'openbsd')):
+                            if self._opts.tmux and os.getenv('TMUX_PANE'):
+                                copier_params = 'tmux_detected'
+                            else:
```
jarun: please change it this way:

```
copier_params = []

if linux-ish et al platform:
    if shutil.which('xsel')
        copier_params = ['xsel', '-b', '-i']
    elif 'TMUX_PANE` available:
        copier_params = ['tmux_detected'] // or append, anything
elif elif sys.platform == 'darwin':
    ...
elif elif sys.platform == 'win32':
    ...
...
```

jarun: Check the final code from here as I have updated it.

zmwangx:
Give me a few minutes, I'll do it instead...

jarun:
Sure! If you can test the changes, please go ahead.

jarun:
I probably lost interest in this as soon as I saw tmux in the original issue. it's probably the slowest and most obfuscated piece of crapware I have ever used. I tried it once. Nothing worked out of the box for me. So sorry about my general lack of interest.

zmwangx:
Re tmux: it's not great, but sometimes it's the best you've got. I have 18 panes at the moment. Tmux control mode with a good terminal emulator that supports it (i.e., use your terminal emulator's native windowing, paning, etc.) really take most of the pain out of it.

jarun:
I seldom work remotely nowadays. And multi-tabbed terminator/xfce4-terminal/AltYo in drop-down mode are the poisons for me. ;)

jarun:
Closing this as PR #232 is a better way to handle.

jose1711:
the idea of original PR was that user still has some control of whether the url will be sent to X11 clipboard (when running tmux inside xterm for instance) or passed to tmux. current implementation takes this option away and if tmux is detected it uses tmux'es buffer.

zmwangx:
Unfortunately, we're just averse to adding even more options. Since googler's just a single file and pretty much in feature-freeze at the moment, it shouldn't be hard to maintain a "fork" and have it your way.

jose1711:
yes, you're right. i am thinking.. a compromise could be trying if ```xsel``` won't fail due to undefined ```DISPLAY```. user could then point ```googler``` to X11/tmux by (un)setting ```DISPLAY```. so code-wise: try to execute ```xsel```. if it fails continue detecting tmux pane. what do you think?

jarun:
I believe a much safer (and less complicated) bet is an environment variable `FORCE_TMUX_BUFFER`.

However, given googler's age and popularity and the fact this is the first defect related to `tmux` we are OK with the current status. `xsel` is not even installed on Ubuntu by default (surprise!). If you don't want it, just don't install it or remove it if it is installed in your system. It's too early to put more effort into it.

PR #237, opened 2018-06-05
Update README.md

TonyStark:
Installation instruction in Android

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,18 @@ and upgrade by running
 
     $ sudo googler -u --include-git
 
+### Installation in Android
+
+- First Install `Termux` app from Playstore
+- Open it and install python and curl by following command
+
+`pkg install python python-dev curl`
+- Then download googler and give proper permission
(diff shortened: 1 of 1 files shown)


zmwangx:
I'd say no to this. I've never seen a comparable project with a README that includes installation instructions for Android, or HP-UX, or Solaris, or whatever. Also, it would be weird for us to endorse a specific terminal emulator from Play Store, out of all places.

jarun:
Oh come on! This is from Tony Stark, the Iron Man himself!

Jokes apart, your argument makes sense. So I'll save it in a gist or somewhere.

kenorb:
You can add as a Wiki page.

jarun:
> You can add as a Wiki page.

That's a good idea. I'll add it.

jarun:
Added here - https://github.com/jarun/googler/wiki/Unofficial-guide-to-Android-installation

Thank you!

PR #243, opened 2018-08-09
Add tags to gitignore

RatanRSur:
In case you don't use it, many people use ctags to jump to definitions of classes/functions etc. Thought I'd add this while I was at it :)


RatanRSur:
Ah yes, I forgot about that.

PR #245, opened 2018-09-21
Unescape html for query auto completions

geniusgordon:
Hi,
The results from the auto-completion contain html escaped entities.

```
$ googler --complete what
whatsapp
what&#39;s up # should be what's up
```

This commit uses the built-in html module to unescape it.


jarun:
Thank you!

PR #246, opened 2018-09-30
Update zsh completion for googler

mkjmdski:
When closing free text inside _googler function you can easily just source the file from any place to get the completion functionality. I found it really helpful when using googler from inside the zplug.

### DIFF
--- a/auto-completion/zsh/_googler
+++ b/auto-completion/zsh/_googler
@@ -32,32 +32,34 @@ _googler_complete_query () {
     compadd $@ -- $completions
 }
 
-local -a args
-args=(
-    '(- : *)'{-h,--help}'[show help text and exit]'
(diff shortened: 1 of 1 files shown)


mkjmdski:
Thanks for the explanation @zmwangx :). That is really helpful.

zmwangx:
No problem. More detailed explanation of autoloaded functions in Zsh, if you're unfamiliar with the concept: http://zsh.sourceforge.net/Doc/Release/Functions.html#Autoloading-Functions. Also, check out http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Autoloaded-files (or `man zshcompsys`) for autoloading in the context of the completion system, how the `#compdef` tag works, how `compinit`, `compaudit` (you'll need to deal with this when directories in `$fpath` has the wrong permissions) and co. operate, etc.

PR #252, opened 2018-10-18
[WIP] Parser rewrite

zmwangx:
The current version is already feature-complete (I believe).

Some highlights:

- Parser logic is <100 loc now.
- A new option `-D, --debugger` to open an IPython/Python shell to interactively inspect the DOM tree. Good for developing/debugging the parser in the case of upstream changes.
- "No results found for ... Results for ...:" now handled.

### DIFF
--- a/googler
+++ b/googler
@@ -16,6 +16,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+# TODO: showing results for... in --noprompt mode
+
 import argparse
 import atexit
 import base64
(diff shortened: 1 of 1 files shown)


jarun:
Awesome! I will review it this weekend.

jarun:
Looks good to me. Thank you!

jarun:
Please raise a new PR for the `--noprompt` ToDo.

zmwangx:
Actually, I've yet to add new testing code, which is one of the reasons why this was marked as WIP ;) A new test framework and suite is in the pipeline.

zmwangx:
Oh, let me jot down one potential concern before I forget.

As I said, the introduced `dim` module is strictly typed, so it has type annotations. However, Python 3.7 introduced postponed evaluation of type annotations[1], which is a good thing that I couldn't take advantage of since I had to support 3.4 to 3.6. This in itself is not an issue, but PEP 536 comes with a deprecation policy:

- In Python 3.8 a `PendingDeprecationWarning` is raised by the compiler in the presence of type annotations in modules without the `__future__` import.
- Starting with Python 3.9 the warning becomes a `DeprecationWarning`.

So, taken at face value, we'll see a `PendingDeprecationWarning` with googler when Python 3.8 comes out.

I'm not overly concerned about this though, nor have I given this much thought, since this is pretty big and will affect all projects using type annotations and supporting both 3.6 and 3.8 -- that's a lot of projects. (We haven't seen a `__future__` feature this big since py3k[3]. The only Python 3.x `__future__` feature before this was `generator_stop`, and it was a pretty niche feature.) I'm confident something will be worked out.  Worst case scenario, we can strip all the type annotations; after all, `dim` has an independent repo that's type-checked, and the googler repo isn't type-checked.

[1] https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-pep563
[2] https://www.python.org/dev/peps/pep-0563/#deprecation-policy
[3] https://docs.python.org/3/library/__future__.html

jarun:
> I've yet to add new testing code

I've been using it for a few days now. ;)

PR #255, opened 2018-10-23
ISSUE_TEMPLATE: try to emphasize reproduction on master

zmwangx:


jarun:
bombs ;)

zmwangx:
If there's a more eye-catching symbol that's safe for work, I'm all ears ;)

PR #263, opened 2018-11-27
Rewrite URL copying with screen buffer

zmwangx:
I was reminded tonight, by chance, of this feature, which I intended to rewrite when I first saw it, but then forgot.

---

A couple of issues:

- Avoid race condition with a proper mkstemp'ed file. The primary purpose of
  this commit.

- The copier mode shouldn't be called `ext_file` because it sounds like a
  generic writing to then reading from an external file (btw, what's an
  internal file?), when it's really just GNU Screen-specific. So call it what
  it is.

- Enumerate all possible values of copier_mode, then leave one final
  theoretically unreachable else clause. Makes the final case easier to
  understand without reading a lot of prior code. Like switch/case/default. (I
  actually would like a switch in Python, but PEP 3103 was rejected due to "A
  quick poll during my keynote presentation at PyCon 2007 shows this proposal
  has no popular support. I therefore reject it." I digress.)

### DIFF
--- a/googler
+++ b/googler
@@ -2744,9 +2744,9 @@ class GooglerCmd(object):
                                 copier_params = ['tmux', 'set-buffer']
                                 copier_mode = 'cmdline_arg'
                             elif os.getenv('STY'):  # check for GNU screen
-                                # Try to use GNU screen's exchange-file as fallback.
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #269, opened 2019-02-16
Update TLS hardening code for py37 w/ OpenSSL 1.1

zmwangx:
- Use `SSLContext.minimum_version` for py37 w/ OpenSSL 1.1.0g+. (`OP_NO_*` options have been deprecated going forward.) [1]

- Rename `TLS1_2Connection` to `HardenedHTTPSConnection`. `TLS1_2Connection` is a bad name since the class doesn't always negotiate TLS 1.2, and TLS 1.2 isn't the final version of TLS.

  (Apparently I have TLS 1.3 in mind, which improves latency by saving a roundtrip.)

[1] https://docs.python.org/3/library/ssl.html#ssl.OP_NO_TLSv1

### DIFF
--- a/googler
+++ b/googler
@@ -1424,7 +1424,7 @@ def check_stdout_encoding():
 
 # Classes
 
-class TLS1_2Connection(HTTPSConnection):
+class HardenedHTTPSConnection(HTTPSConnection):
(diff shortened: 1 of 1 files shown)


jarun:
Thank you! :+1:

PR #272, opened 2019-03-27
Enable result metadata extraction in Search mode

zmwangx:
Result metadata (contained in a div.slp in search results) was originally
designed to be extracted in News mode, for publisher name and time of
publishing. (In Search mode, the most important metadata -- date of the page,
when available -- is already contained in the abstract itself. Maybe this info
could be promoted to the separate metadata field, but that's the topic for
another day.)

However, the old parser did not explicitly restrict the extraction to News
mode, which unintentionally let through useful information, e.g., IMDb rating
when searching for a movie. The new parser (ad965d6) instead restricted
metadata extraction to News mode, causing a regression in certain cases.

This commit removes the limitation and allows metadata extraction whenever
possible.

### DIFF
--- a/googler
+++ b/googler
@@ -2112,10 +2112,10 @@ class GoogleParser(object):
                     title = mime.text + ' ' + title
                 url = self.unwrap_link(a.attr('href'))
                 abstract = div_g.select('.st').text.replace('\n', '')
-                if self.news:
+                try:
                     metadata = div_g.select('.slp').text
-                    metadata = metadata.replace('\u200e', '').replace(' - ', ', ')
-                else:
+                    metadata = metadata.replace('\u200e', '').replace(' - ', ', ').strip()
+                except AttributeError:
                     metadata = None
             except (AttributeError, ValueError):
                 continue


zmwangx:
There's one caveat: Note that in order to make metadata of news entries look nicer, we do this transformation on metadata text:

```py
metadata = metadata.replace('\u200e', '').replace(' - ', ', ').strip()  # .strip() is new but inconsequential
```

However, when we search for a movie, this is what happens:

```console
$ googler -n 1 --np avatar

 1.  Avatar (2009) - IMDb
     https://www.imdb.com/title/tt0499549/
     Rating: 7.8/10 - 1,032,269 votes
     Photos. Sam Worthington in Avatar (2009) Wes Studi in Avatar (2009) Avatar prop in the Profiles in History booth Sigourney Weaver at an event for
     Avatar (2009) ...
```

Why isn't "Rating: 7.8/10 - 1,032,269 votes" transformed into "Rating: 7.8/10, 1,032,269 votes" (less readable)? Interestingly, with our `googler/*` user agent, this is the actual HTML of the `div.slp`:

```html
<div class="f slp"><div class="star"><div style="width:52px">&nbsp;</div></div>&nbsp;Rating: 7.8/10&nbsp;-&nbsp;1,032,269 votes</div>
```

so the hyphen is surrounded by &nbsp;'s (i.e., U+00A0) instead of spaces; that's why the transform doesn't work (happy coincidence). However, when I search in Chrome, this is the HTML for the `div.slp`:

```html
<div class="slp f"><g-review-stars><span class="fTKmHE99XE4__star-default" aria-label="Rated 4.0 out of 5,"><span style="width:53px"></span></span></g-review-stars> Rating: 7.8/10 - ‎1,032,269 votes</div>
```

so if we were parsing this HTML instead, the undesirable transform would kick in, and there may be a future date (or maybe even a Google variant today) when it kicks in for our user agent too.

Anyway, not that big a deal, just documenting this in case of future confusion. (Basically, to future self: I thought about this.)

jarun:
Thank you!

zmwangx:
Screen Shot 2019-03-27 at 10 42 54 PM

Looking at the email, for a second there I thought you recruited a new maintainer.

jarun:
LOL! No just made it a little more difficult for people to figure me out
right away ;)

Regards,

Arun

PR #275, opened 2019-04-09
Enable/Disable ANSI Sequences on Windows.

guilt:
Similar to https://github.com/jarun/ddgr/pull/87/

### DIFF
--- a/googler
+++ b/googler
@@ -30,6 +30,7 @@ from http.client import HTTPSConnection
 import locale
 import logging
 import os
+import platform
 import shutil
 import signal
 import socket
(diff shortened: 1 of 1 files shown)


jarun:
@zmwangx @dertuxmalwieder can you guys please review this PR?

dertuxmalwieder:
Setting use_colors up to three times looks inefficient. It might work though. Is it designed to work with both cmd and ConEmu? (I remember having patched such for buku.)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jarun:
@guilt please take a look at the relevant Buku code.

dertuxmalwieder:
Discussed here: https://github.com/jarun/Buku/issues/217

guilt:
This is only designed to work with standard cmd.

jarun:
@guilt does the patch mentioned above work for you?

guilt:
The patch above should work. But yes, what it does is disable ANSI in standard Command Prompt and only enable it in ConEmu.

This one does not do ConEmu but selectively enables ANSI sequences on Command Prompt (which was basically the main problem child, warranting ConEmu/what not).

If there's a detection for ConEmu, I'd want someone to look into support for a good % emulated environments on Windows.

guilt:
Anyway, my patch works for me. I'm removing my fork and keep it his PR as reference. Can't spend more time on it. Sorry.

zmwangx:
Seems okay functionally, but code isn't very understandable (quite a few magic numbers) and I'd be cautious to include code I don't understand. It's definitely editing code at the wrong place though — the action for `--nocolor` apparently has to be `store_false`; otherwise, if your `use_colors` happen to be false, `--nocolor` would actually enable color. So the PR needs a medium rewrite anyway.

Since you don't want to spend more time on this, I guess I'll rewrite it later while cross referencing colorama's win32 implementation. A brief look tells me that you might have only modified stdout (where the magic number -11 comes from) but not stderr.

zmwangx:
Update: please ignore the "wrong place" part of the previous comment. I only glanced at the code and somehow had the wrong impression.

jarun:
@guilt no problem!

@zmwangx is it fine to close this PR then?

PR #278, opened 2019-04-10
Better explain "Connection blocked due to unusual activity."

zmwangx:
Before:

```console
$ googler "I use googler for automated scraping."
[ERROR] Connection blocked due to unusual activity.
```

After:

```console
[ERROR] Connection blocked due to unusual activity. THIS IS NOT A BUG, please do NOT report it as a bug unless you have specific information that may lead to the development of a workaround. You IP address is temporarily or permanently blocked by Google and requires reCAPTCHA-solving to use the service, which googler is not capable of. Possible causes include issuing too many queries in a short time frame, or operating from a shared / low reputation IP with a history of abuse. Please do NOT use googler for automated scraping.
```


jarun:
Thank you!

PR #279, opened 2019-04-11
Drop tests/ci-test-wrapper

zmwangx:
ci-test-wrapper is a relic of the past that's not necessary anymore. We can
now comfortably test all commits including README typo fixes.

Also in this commit:

- Add NUM_TEST_ITERATIONS and SLEEP_DURATION to .circleci/config.yml so that
  secret environment variables are only used for actual secrets and don't need
  to be shared with PR builds.

- Trap SIGINT in tests/test to make interrupting the script easier.


jarun:
Thank you!

PR #280, opened 2019-04-11
win32: also ENABLE_VIRTUAL_TERMINAL_PROCESSING on stderr

zmwangx:
Follow-up on #277.

As discussed here[1], it's not really necessary when stdout and stderr are the
same console, but it doesn't hurt either.

Note that we did not test sys.stderr.isatty() in the code. This has been
considered, and deemed not a problem, since stdout is handled first (what
really matters) and we have a catch-all pass for any possible exception.

[1] https://github.com/jarun/googler/pull/277#discussion_r274245020

### DIFF
--- a/googler
+++ b/googler
@@ -3247,17 +3247,19 @@ def main():
             # https://docs.microsoft.com/en-us/windows/console/setconsolemode
             if platform.release() == '10':
                 STD_OUTPUT_HANDLE = -11
+                STD_ERROR_HANDLE = -12
                 ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004
                 try:
                     from ctypes import windll, wintypes, byref
                     kernel32 = windll.kernel32
-                    stdout_handle = kernel32.GetStdHandle(STD_OUTPUT_HANDLE)
-                    old_mode = wintypes.DWORD()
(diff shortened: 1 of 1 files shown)


zmwangx:
Nice number by the way.

Screen Shot 2019-04-11 at 1 08 55 PM

guilt:
We're not considering the case where stdout goes to file and stderr does not :) But yeah, that's not a general usecase, so we should be good.

In general, we should be getting the same handle for STDOUT and STDERR, so we're good.

zmwangx:
stdout goes to file => no color by default.

stdout goes to file, color enabled by user, user still wants to use interactive prompt for whatever reason (errors are not colored) => user is insane.

googler
```diff
@@ -3247,17 +3247,19 @@ def main():
             # https://docs.microsoft.com/en-us/windows/console/setconsolemode
             if platform.release() == '10':
                 STD_OUTPUT_HANDLE = -11
+                STD_ERROR_HANDLE = -12
```
jarun: Can we put this complete block (from line 3245 to 3266) within a separate API?

jarun:
Oops, I think I've already merged it. But anyway, can we do the API separation?

PR #283, opened 2019-05-07
Show matched keywords as bold in result abstracts

EvanDotPro:
Useful for telling when Google has matched synonyms to your query (for example,
search "stationary supplies" or "aosp"). This is very helpful information for
deciding when to wrap a word or phrase in quotes to force an exact match, or
otherwise refine your search query.

- Does not bolden matched keywords if colors are disabled.
- Includes array of matched keyword for each result in json output.
- Will look into tests upon receiving feedback on the idea.

### Examples:

Screenshot from 2019-05-07 11-02-42

Screenshot from 2019-05-07 11-14-14

### DIFF
--- a/googler
+++ b/googler
@@ -2121,7 +2121,12 @@ class GoogleParser(object):
                 if mime:
                     title = mime.text + ' ' + title
                 url = self.unwrap_link(a.attr('href'))
-                abstract = div_g.select('.st').text.replace('\n', '')
+                matched_keywords = []
+                abstract = ''
+                for childnode in div_g.select('.st').children:
+                    if childnode.tag == 'b' and childnode.text != '...':
+                            matched_keywords.append({'phrase': childnode.text, 'offset': len(abstract)})
+                    abstract = abstract + childnode.text.replace('\n', '')
                 try:
                     metadata = div_g.select('.slp').text
                     metadata = metadata.replace('\u200e', '').replace(' - ', ', ').strip()
@@ -2141,7 +2146,7 @@ class GoogleParser(object):
                     continue
             index += 1
             self.results.append(Result(index, title, url, abstract,
-                                       metadata=metadata, sitelinks=sitelinks))
+                                       metadata=metadata, sitelinks=sitelinks, matches=matched_keywords))
 
         # Showing results for ...
         # Search instead for ...
@@ -2221,6 +2226,7 @@ class Result(object):
     abstract : str
     metadata : str or None
     sitelinks : list
+    matches : list
 
     Class Variables
     ---------------
@@ -2238,7 +2244,7 @@ class Result(object):
     colors = None
     urlexpand = True
 
-    def __init__(self, index, title, url, abstract, metadata=None, sitelinks=None):
+    def __init__(self, index, title, url, abstract, metadata=None, sitelinks=None, matches=None):
         index = str(index)
         self.index = index
         self.title = title
@@ -2246,6 +2252,7 @@ class Result(object):
         self.abstract = abstract
         self.metadata = metadata
         self.sitelinks = [] if sitelinks is None else sitelinks
+        self.matches = [] if matches is None else matches
 
         self._urltable = {index: url}
         subindex = 'a'
@@ -2276,7 +2283,7 @@ class Result(object):
             else:
                 print(' %s%-*s %s %s' % (' ' * pre, indent, index + '.', title, url))
 
-    def _print_metadata_and_abstract(self, abstract, metadata=None, indent=5, pre=0):
+    def _print_metadata_and_abstract(self, abstract, metadata=None, matches=None, indent=5, pre=0):
         colors = self.colors
         try:
             columns, _ = os.get_terminal_size()
@@ -2290,6 +2297,15 @@ class Result(object):
                 print(' ' * (indent + pre) + metadata)
 
         if colors:
+            # Start from the last match, as inserting the bold characters changes the offsets.
+            for match in reversed(matches or []):
+                abstract = (
+                    abstract[: match['offset']]
+                    + '\033[1m'
+                    + match['phrase']
+                    + '\033[0m'
+                    + abstract[match['offset'] + len(match['phrase']) :]
+                )
             print(colors.abstract, end='')
         if columns > indent + 1 + pre:
             # Try to fill to columns
@@ -2305,7 +2321,7 @@ class Result(object):
     def print(self):
         """Print the result entry."""
         self._print_title_and_url(self.index, self.title, self.url)
-        self._print_metadata_and_abstract(self.abstract, metadata=self.metadata)
+        self._print_metadata_and_abstract(self.abstract, metadata=self.metadata, matches=self.matches)
 
         for sitelink in self.sitelinks:
             self._print_title_and_url(sitelink.index, sitelink.title, sitelink.url, pre=4)
@@ -2322,6 +2338,8 @@ class Result(object):
             obj['metadata'] = self.metadata
         if self.sitelinks:
             obj['sitelinks'] = [sitelink.__dict__ for sitelink in self.sitelinks]
+        if self.matches:
+            obj['matches'] = self.matches
         return obj
 
     def urltable(self):


jarun:
This is really cool and I was kind of hoping someone contributes this. Thank you!

> Includes array of matched keyword for each result in json output.

Can you share an example output for a visual cue?

EvanDotPro:
@jarun Just pushed a fix to the json bit (went a little copy/paste happy there). Also added an example screenshot up in the PR description.

jarun:
Thank you!

@zmwangx please take a look.

EvanDotPro:
Just noticed COLORMAP with 'X' and 'x' defined for bold/normal already. I'll update to use that in a bit.

jarun:
Sure!

jarun:
A quick note, beware of the cases where the length of the keyword is very short e.g. 12... I think you need to ensure we highlight words only and not substrings. Another thing would be to ensure the mechanism works even when the word is at the beginning or end.

EvanDotPro:
I believe those cases (substring, beginning, and end) are all covered by my use of [\b](https://www.regular-expressions.info/wordboundaries.html) in the regex replacement. I did quite a bit of local testing, but if you find any cases where `\b` is not sufficient, let me know and I'll update the PR accordingly.

jarun:
Sure!

jarun:
Once you are done with the COLORMAP X use, leave a note.

zmwangx:
Interactive part: LGTM. Don't bother with using `COLORMAP`, it's there for mapping user-defined color themes to actual sequenced. Fixed sequences can be hard-coded, no problem.

Parser and structured output changes: not so sure. Does anyone actually need that? If one actually does need that, just the matching part seems be incomplete info, too — position of a match is almost as important as the text of the match (think of information captured by a `re.Match`). Lastly, if someone needs this, they can simply do a regex query themselves like you've shown and get all the information.

I'd say let's just not bother with the addition to structured output.

googler
```diff
@@ -2277,7 +2286,7 @@ class Result(object):
             else:
                 print(' %s%-*s %s %s' % (' ' * pre, indent, index + '.', title, url))
 
-    def _print_metadata_and_abstract(self, abstract, metadata=None, indent=5, pre=0):
+    def _print_metadata_and_abstract(self, abstract, metadata=None, keywords=[], indent=5, pre=0):
```
zmwangx: Use `None` for the default value. Don't use a mutable default value for a keyword argument. See for instance https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.

zmwangx: (I know, we don't actually mutate this specific argument in the method body, but it's code smell.)

zmwangx:
An update to my last comment: sorry, I didn't read the code in detail and misunderstood what's being matched. You're matching against matched phrases highlighted in Google's response, not phrases supplied by the user (what I thought). In that case, the parser addition is definitely warranted. Still not that convinced about addition to structured output, though.

zmwangx:
Just came up with an important use case(s) where the re matching strategy fails: any written language that doesn't have visible word boundaries. e.g., Chinese or Japanese. Try e.g. `googler 中文`; nothing is highlighted. So, while regex matching is smart, it has crippling limitations after all. My initial response is fitting:

> position of a match is almost as important as the text of the match

And I guess the information has to be recorded in the parser, not afterwards with clever tricks.

(Not trying to diminish this feature in any way.)

EvanDotPro:
Yeah, the whole idea here is that often Google matches words and phrases that it feels are synonymous with your search terms, and it's useful to the user to know when Google is doing this sort of loose matching.

When you say structured output, are you talking about JSON?

zmwangx:
> it feels are synonymous with your search terms

Makes sense.

> When you say structured output, are you talking about JSON?

Yeah.

EvanDotPro:
Well, as one example, I have a personal side project in mind where I'll be consuming the JSON for a non-cli interface, which I will need to know which terms Google has matched in order to make them bold in the UI (for the same reasons we'd want them bold in the CLI interface).

zmwangx:
Actually, I was already convinced matches could/should be added to structured output when I had [this realization](https://github.com/jarun/googler/pull/283#issuecomment-490321282). But as I said, match positions also have to be included, otherwise we can only support a subset of languages. Since schema changes should be done with extreme care, it has to be all or nothing.

zmwangx:
In case I wasn't clear enough, what I have in mind is something like:

```json
{
  "url": "...",
  "title": "...",
  "abstract": "...",
  "matches": [
    {"phrase": "...", "offset": ...},
    {"phrase": "...", "offset": ...},
    ...
  ]
}
```

EvanDotPro:
Totally understood. Hadn't given any thought as to how regex's \b handles unicode, my bad. I'll hopefully have time to get these changes in tomorrow. :+1:

zmwangx:
No problem, this is not at all obvious for people who are only familiar with the Latin/Cyrillic/etc. writing systems. Consider this simple Chinese phrase, "中文书面语", which means "Chinese written language"; it can be broken down into two or three words: "中文" — Chinese, "书面语" — written language (which could be further broken into "书面" — written, "语" — language), but there's no word boundary character inside it:

```py
>>> re.findall(r'\b.', '中文书面语')
['中']
```

(As a result tokenization of Chinese/Japanese text is nontrivial and requires specialized tokenizers, a simple example being ICU's BreakIterator.)

As for implementation, iterating over children of a `div_g.select('.st')` and manually build up the text content (instead of using the `text` property) while recording positions as needed should do the trick. The HTML parser's API is documented at https://docs.tcl.sh/py/dim/ in case you need it.

jarun:
@EvanDotPro any update on this?

EvanDotPro:
Sorry for the delay on this @jarun, been very busy the last couple weeks.

@zmwangx I implemented your suggestions for improved support of non-English unicode words. Tested it with your example above and it seems to be working as expected.

Screenshot from 2019-05-14 10-06-14

I've updated the JSON schema as suggested as well:

Screenshot from 2019-05-14 10-07-11

jarun:
@zmwangx please review.

googler
```diff
@@ -2277,7 +2284,7 @@ class Result(object):
             else:
                 print(' %s%-*s %s %s' % (' ' * pre, indent, index + '.', title, url))
 
-    def _print_metadata_and_abstract(self, abstract, metadata=None, indent=5, pre=0):
+    def _print_metadata_and_abstract(self, abstract, metadata=None, matches=[], indent=5, pre=0):
```
zmwangx: `matches=None`.

googler
```diff
@@ -2291,6 +2298,10 @@ class Result(object):
                 print(' ' * (indent + pre) + metadata)
 
         if colors:
+            # Start from the last match, as inserting the bold characters changes the offsets.
+            matches.reverse()
+            for match in matches:
```
zmwangx: `for match in reversed(matches)` is enough, and doesn't mutate the argument, which is an unintended side effect.

googler
```diff
@@ -2291,6 +2298,10 @@ class Result(object):
...
+                abstract = abstract[:match['offset']] + '\033[1m' + match['phrase'] + '\033[0m' + abstract[match['offset']+len(match['phrase']):]
```
zmwangx: Line too long, break it up; also, for consistency, use spaces around operators. Below is black's take on this statement, with quotes replaced. (I'm not a fan of black, and it's obviously too late to blacken googler, but it does stop all arguments on code formatting.)

```py
abstract = (
    abstract[: match['offset']]
    + '\033[1m'
    + match['phrase']
    + '\033[0m'
    + abstract[match['offset'] + len(match['phrase']) :]
)
```

googler
zmwangx: Actually, should use `reversed(matches or [])` now that `matches` could be `None`.

jarun:
@EvanDotPro any update on this? We are planning to make a release soon. I would love to have this in.

jarun:
@zmwangx I guess @EvanDotPro is busy. If we accept this PR in its current state would you be able to finish it off without spending too much time? I really want to have it in next the release.

zmwangx:
Sorry for the slow reply here... I believe I already thoroughly reviewed the changed last time and left very specific actionable comments. I'll fix up the commit according to my comments tomorrow, do a bit of testing, and close this.

jarun:
Thank you!

zmwangx:
Fixed up the review comments, squashed (author date reset, author kept) and force pushed. Will merge into master when green.

jarun:
Thank you!

PR #303, opened 2019-11-16
Point user to wiki article when there are no results

zmwangx:
I wrote a wiki article about what to do when "No results" happens: https://git.io/googler-no-results. Hopefully it will reduce user anxiety and increase efficiency, and save some explanations when the inevitable happens again.

The message is shown at most once per session. It is not shown in --json mode, obviously.

Example:

    $ googler adjskfasdfhasjkfhsajkhdfa
    No results.
    If you believe this is a bug, please review https://git.io/googler-no-results before submitting a bug report.
    googler (? for help) asdfasdfhajsdfhasjkdfhajshdfahjdsk
    No results.
    googler (? for help)

### DIFF
--- a/googler
+++ b/googler
@@ -2524,6 +2524,8 @@ class GooglerCmd(object):
 
         self.promptcolor = True if os.getenv('DISABLE_PROMPT_COLOR') is None else False
 
+        self.no_results_instructions_shown = False
+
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #300, opened 2019-11-13
Improve textwrap in presence of zero-width sequences

zmwangx:
Fixes #287 which I promised to fix long ago.

Three commits: 90f4ea4 is some cleanup work around indentation I committed back in June... I re-examined it and it should be good. See commit message for details. Next is some py34 legacy removal paving the way for today's work. Then there's the meat of this PR. I'll repeat the commit message here.

---
    
Example (with keyword highlighting on, but the highlight effect is
obviously lost in textual form):

Before:

```console
$ googler -n3 --np linux                                                       |
                                                                               |
 1.  Linux.org                                                                 |
     https://www.linux.org/                                                    |
     5 days ago ... Friendly Linux Forum. ... This is a video from my          |
     series of chapters in my book "Essential Linux Command Line"...           |
     Continue… Load more…                                                      |
                                                                               |
 2.  Linux - Wikipedia                                                         |
     https://en.wikipedia.org/wiki/Linux                                       |
     Linux is a family of open source Unix-like operating systems              |
     based on the Linux kernel, an operating system kernel first               |
     released on September 17, 1991, by ...                                    |
                                                                               |
 3.  The Linux Foundation – Supporting Open Source Ecosystems                  |
     https://www.linuxfoundation.org/                                          |
     The Linux Foundation supports the creation of sustainable open            |
     source projects and ecosystems in blockchain, deep learning, networking,  |
     and more.                                                                 |
                                                                               |
```

After:

```console
$ googler -n3 --np linux                                                       |
                                                                               |
 1.  Linux.org                                                                 |
     https://www.linux.org/                                                    |
     5 days ago ... Friendly Linux Forum. ... This is a video from my series   |
     of chapters in my book "Essential Linux Command Line"... Continue… Load   |
     more…                                                                     |
                                                                               |
 2.  Linux - Wikipedia                                                         |
     https://en.wikipedia.org/wiki/Linux                                       |
     Linux is a family of open source Unix-like operating systems based on the |
     Linux kernel, an operating system kernel first released on September 17,  |
     1991, by ...                                                              |
                                                                               |
 3.  The Linux Foundation – Supporting Open Source Ecosystems                  |
     https://www.linuxfoundation.org/                                          |
     The Linux Foundation supports the creation of sustainable open source     |
     projects and ecosystems in blockchain, deep learning, networking, and     |
     more.                                                                     |
                                                                               |
```

The idea is to use a text wrapper that keeps track of the position of
each source character, so that zero-width sequences can be inserted at
[...]

PR #304, opened 2019-11-16
Drop option --nohl

zmwangx:
This reverts commit 7e725573f801c587df7688a07f9fc7a9408d76eb.

Per discussion[1], the value proposition has changed now that text wrapping in
presence of ANSI escape sequences has been fixed.

[1] https://github.com/jarun/googler/pull/300#issuecomment-554608255


jarun:
Dropped! ;)

PR #305, opened 2019-11-20
dim: merge in minor correctness fix

zmwangx:
`HTMLParser.handle_starttag` and `HTMLParser.handle_startendtag` now expect
`List[Tuple[str, Optional[str]]`, acknowledging the fact that attribute values
may be None. Our implementation does not expect None, so we convert None to
the empty string.

https://github.com/zmwangx/dim/commit/8d71282d1d036c57385cd2e4310616220076fe3c

### DIFF
--- a/googler
+++ b/googler
@@ -64,6 +64,7 @@ from typing import (
     List,
     Match,
     Optional,
+    Sequence,
     Tuple,
     Union,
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #311, opened 2019-12-10
Fix .replit

amasad:
Repl.it struggles with persisting file permissions so this changes the run command to run it as a python script instead of an executable. cc @Mosrod 
Screen Shot 2019-12-09 at 10 07 35 PM


jarun:
Fixed with PR.

amasad:
@jarun I'm not sure what you mean but it doesn't look fixed

PR #315, opened 2019-12-29
Fixed repl.it command

PDanielY:
@Mosrod did the thing wrong. So I fixed it 👍 

> Also not trying to sound mean or anything

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,2 +1,2 @@
 language = "python3"
-run = "./googler"
+run = "python googler"


jarun:
Can you fix pdd as well? https://github.com/jarun/pdd

PDanielY:
I'll try

PDanielY:
The repl works fine. Did someone say that couldn't run it?

jarun:
OK. So `chmod` does the trick. Thanks!

PR #322, opened 2020-02-24
Introduce options --from and --to for date range

zmwangx:
Fixes #320.

- Note1: I'm not 100% sure the dates must match `r'^(\d+/){0,2}\d+$'`.
- Note2: Not gonna introduce more short options, in part because `--to` clashes with `-t`, and the usual convention of making it `-T` somehow doesn't feel right.

PR #324, opened 2020-03-04
Fix regression caused by #322

zmwangx:
Need to check not None and convert from None to empty string when dealing with
from and to.


jarun:
Thank you!

PR #326, opened 2020-03-24
Add the `--url-output` option for writing selected URLs to a file.

kkysen:
Instead of running a new process for each selected URL like with `--url-handler`, 
`--url-output` takes a file path, which it writes the selected URLs to, delimited by `\n`.

This can be much more useful and efficient than `--url-handler`.
`--url-handler` requires an executable to run for each URL,
but does not allow any extra arguments to be sent to the program,
and since it must be an executable, it can't be a bash function, for example.

`--url-output`, when the file path is a named pipe, allows for much more robust url handling.
The program that reads from the pipe as googler writes to it can do anything,
including running a bash function or passing extra args to it, 
and it also avoids forking a new process for each URL.

### DIFF
--- a/googler
+++ b/googler
@@ -3350,9 +3350,9 @@ def parse_args(args=None, namespace=None):
            help='disable TCP optimizations and forced TLS 1.2')
     addarg('--json', action='store_true',
            help='output in JSON format; implies --noprompt')
-    addarg('--url-handler', metavar='URL_HANDLER',
+    addarg('--url-handler', metavar='UTIL',
            help='custom script or cli utility to open results')
-    addarg('--url-output', metavar='URL_OUTPUT',
+    addarg('--url-output', metavar='FILE',
            help='file to write urls to')
     addarg('--show-browser-logs', action='store_true',
            help='do not suppress browser output (stdout and stderr)')
--- a/googler
+++ b/googler
@@ -1390,6 +1390,13 @@ def open_url(url):
     in a custom CLI script or utility. A subprocess is spawned with url as
     the parameter in this case instead of the usual webbrowser.open() call.
 
+    The file stream ``open_url.url_output`` can be used to write opened URLs to.
+    The URL and a newline are written to the file, which is then flushed.
+    When the file is a pipe, this makes it easy for a shell to read in the URLs
+    from the named pipe and run commands on them in parallel to googler.
+    It is very difficult (and/or slow) to run bash functions, for example,
+    using ``open_url.url_handler``.  This makes that much simpler.
+
     Whether the browser's output (both stdout and stderr) are suppressed
     depends on the boolean attribute ``open_url.suppress_browser_output``.
     If the attribute is not set upon a call, set it to a default value,
@@ -1408,6 +1415,11 @@ def open_url(url):
     if hasattr(open_url, 'url_handler'):
         p = Popen([open_url.url_handler, url], stdin=PIPE)
         p.communicate()
+        return
+    if hasattr(open_url, 'url_output'):
+        out = open_url.url_output
+        out.write(url + '\n')
+        out.flush()
         return
 
     browser = webbrowser.get()
@@ -3338,8 +3350,10 @@ def parse_args(args=None, namespace=None):
            help='disable TCP optimizations and forced TLS 1.2')
     addarg('--json', action='store_true',
            help='output in JSON format; implies --noprompt')
-    addarg('--url-handler', metavar='UTIL',
+    addarg('--url-handler', metavar='URL_HANDLER',
            help='custom script or cli utility to open results')
+    addarg('--url-output', metavar='URL_OUTPUT',
+           help='file to write urls to')
     addarg('--show-browser-logs', action='store_true',
            help='do not suppress browser output (stdout and stderr)')
     addarg('--np', '--noprompt', dest='noninteractive', action='store_true',
@@ -3417,6 +3431,8 @@ def main():
 
         if opts.url_handler is not None:
             open_url.url_handler = opts.url_handler
+        elif opts.url_output is not None:
+            open_url.url_output = open(opts.url_output, 'w')
         else:
             # Set text browser override to False
             open_url.override_text_browser = False


googler
```diff
@@ -3338,8 +3350,10 @@ def parse_args(args=None, namespace=None):
            help='disable TCP optimizations and forced TLS 1.2')
     addarg('--json', action='store_true',
            help='output in JSON format; implies --noprompt')
-    addarg('--url-handler', metavar='UTIL',
+    addarg('--url-handler', metavar='URL_HANDLER',
```
jarun: What's the problem with `UTIl`. This has to be a utility.

googler
```diff
@@ -3338,8 +3350,10 @@ def parse_args(args=None, namespace=None):
...
            help='custom script or cli utility to open results')
+    addarg('--url-output', metavar='URL_OUTPUT',
```
jarun: I think this should be `FILE`.

jarun:
@zmwangx please take a look.

zmwangx:
Apparently this is for interactive use, and merely replaces this handler script

```sh
#!/bin/sh
echo $1 >> /path/to/file
```

Sure, it writes to the file directly instead of spawning <=10 (usually just one) subprocesses. However, if you're that stingy with resources, is something written in Python really the right tool for you? Not convinced.

Meanwhile, to do any postprocessing at all, you at least need another program “listening” on the output file, using it like a poor man's socket, unless I'm misunderstanding something. With such a complicated setup, is it really worth making a two-line script built-in (other than saving a few subprocesses, which as I said I'm not worried about in the least)?

Therefore, my conclusion is I don't recommend adding yet another option that's weirdly specific and likely won't get much adoption outside of one person to our already long list of options, unless a more convincing use case is presented.

googler
kkysen: I didn't realize exactly what you meant by a utility.  I changed it back now.

googler
kkysen: I changed this, too.

jarun:
@kkysen as @zmwangx pointed out, we can have a simple script to pipe the output. It's true that the number of options this tiny utility now has is overwhelming and I agree we shouldn't add more where simple workarounds are available.

kkysen:
I tried doing something like you suggested at first, but the problem is that I can't pass any arguments to the `--url-handler`, i.e., I can't do something like `--url-handler "writeToFile.sh filename"`, so I would have to create a new script for every different file I want to output to.  And to make sure the output file never has any conflicts, it should be a different random file each time, so I'd have to generate this script with the filename hardcoded in before passing it to `--url-handler`.

The other option was to change `--url-handler` to handle multiple arguments by using `[*shlex.split(open_url.url_handler), url]` instead of just `[open_url.url_handler, url]`.  This is a breaking change if someone has a script with a space in its name for some reason, though, so I didn't want to do this.  I could add another `--multiarg-url-handler` option instead, but I thought, if I'm going to add an option anyways, why not just add `--url-output`, which can be used to implement `--multiarg-url-handler` as well.  If you'd rather implement it this way instead, I can change it to either use something like `--multiarg-url-handler` or add the functionality directly to `--url-handler` with the breaking change.

jarun:
@kkysen `googler` isn't going to change much going forward. So just keep your fork and rebase once in 6 months (hopefully).

kkysen:
Ok, if you're not planning on making many changes I guess I can just use my fork instead.  I understand if you don't want to make this change, but I just don't see why adding another option that is only a few extra lines is so bad.

zmwangx:
> so I would have to create a new script for every different file I want to output to.

Not really. Since you want to use a different file (assuming dynamically generated; I can't see another way) each time, you must be in a script setting. In which case you have at least two ways to communicate your dynamic file path to the handler.

## Dynamically generated handler script

Just use a dynamically generated handler script each time. Demo:

```sh
#!/bin/sh
tmpfile="$(mktemp)"
handler="$(mktemp)"
cat >"$handler" <<EOF
#!/bin/sh
echo "\$1" >> "$tmpfile"
EOF
chmod +x "$handler"

# Poor man's coprocess.

coprocess () {
    exec <"$tmpfile"
    while true; do
        IFS= read -r line
        if [ -z "$line" ]; then
            sleep 0.1
        elif [ "$line" = $ ]; then
            break
        else
            echo "handling $line"
        fi
    done
    exec <&-
}

coprocess &
coprocess_pid=$!

googler --url-handler "$handler" "$@"
# Write a sentinel to signal EOF.
echo $ >> "$tmpfile"

wait $coprocess_pid
rm -f "$handler" "$tmpfile"
```

Sample session:

```
/tmp/demo -x googler

 1.  Googler - Wiktionary
     https://en.wiktionary.org/wiki/Googler
     EnglishEdit. EtymologyEdit · Google +‎ -er. NounEdit. Googler (plural Googlers). A full-time Google corporation employee. A regular or habitual user of
     the ...

 2.  jarun/googler: Google from the terminal - GitHub
     googler is a power tool to Google (Web & News) and Google Site Search from the command-line. It shows the title, URL and abstract for each result,
     which can ...

 3.  Googler - Urban Dictionary
     https://www.urbandictionary.com/define.php?term=Googler
     An employee working at Google. Employee benefits include free massages, gourmet food, no set working hours, constant talks from presidential candidates
     and ...

 4.  Googler defined - Googler dictionary definition
     https://www.yourdictionary.com/googler
     Googler definition: (1) A person who is an employee of Google. See Xoogler. (2) A person who uses Google to search the Internet. See Google. ...

 5.  How we hire - Google Careers
     https://careers.google.com/how-we-hire/


 6.  Googlers | Google Blog - The Keyword
     https://www.blog.google/inside-google/googlers/
     How one Googler is raising awareness around ALS · Feb 11 / The She Word ... A Googler's illustrated guide to teamwork · Nov 2019 / Working at
     Google ...

 7.  19 words only Googlers understand - Business Insider
     https://www.businessinsider.com/words-only-googlers-understand-2018-1
     Jan 23, 2018,
     Here are 19 words only Googlers will understand: ... Greyglers: Another Googler nickname, Greyglers are Google employees 40 and older ...

 8.  Guide: Create an employee-to-employee learning ... - re:Work
     https://rework.withgoogle.com/guides/learning-development-employee-to-employee/
     ... through an employee-to-employee network called “g2g” (Googler-to-Googler). ... thousands of Googlers went through an Android training bootcamp run
     by the ...

 9.  Working at Google? What does it take to be a Googler?
     https://medium.com/@dinukajaya/while-i-wait-for-my-google-employment-4423f3ce50d9
     Mar 11, 2016,
     Google undoubtedly is one of the coolest places to work at if you are a tech enthusiast, futurist, idea maniac, business fanatic or a simple soul ...

 10. Not a Googler: The Secret, Frustrating Life of a Google ...
     https://fortune.com/2019/11/07/google-googler-tvc-contract-worker/
     Nov 7, 2019,
     More than half of Google's workers are temporary, vendor or contract staff, known as TVCs.

googler (? for help) o 1-5
googler (? for help) handling https://en.wiktionary.org/wiki/Googler
handling https://github.com/jarun/googler
handling https://www.urbandictionary.com/define.php?term=Googler
handling https://www.yourdictionary.com/googler
handling https://careers.google.com/how-we-hire/
n

 1.  googler — Homebrew Formulae
     https://formulae.brew.sh/formula/googler
     googler. Google Search and News from the command-line. https://github.com/jarun/googler · /api/formula/googler.json (JSON API) · Formula code on
     GitHub.

 2.  Package 'googler' - CRAN
     https://cran.r-project.org/web/packages/googler/googler.pdf
     Package 'googler'. September 4, 2019. Title Google from the R Console. Version 0.0.1. Description This is a wrapper for the command line tool
     'googler', which ...

 3.  Package googler - CRAN
     https://cran.r-project.org/package=googler
     Sep 4, 2019,
     googler: Google from the R Console. This is a wrapper for the command line tool 'googler', which can be found at the following URL: ...

 4.  Coffee with a Googler - YouTube
     https://www.youtube.com/playlist?list=PLOU2XLYxmsIJP13VD_Cg8qS5g2bKWTaYx
     Coffee with a Googler. Google Developers; 64 videos; 117,122 views; Last updated on Oct 24, 2018. Play all. Share. Loading... Save ...

 5.  Install Googler on CentOS using the Snap Store | Snapcraft
     https://snapcraft.io/install/googler/centos
     Feb 10, 2020,
     googler is a power tool to Google (Web & News) and Google Site Search from the command-line. It shows the title, URL and abstract for each ...

 6.  Googler – Google Search from your Linux Terminal | FOSS ...
     https://www.fosslinux.com/25334/googler-google-search-from-your-linux-terminal.htm
     Dec 22, 2019,
     Googler is a powerful tool to perform Google search from the command-line. Let's check its installation and usage.

 7.  Chocolatey Software | googler 3.9
     https://chocolatey.org/packages/googler
     To install googler, run the following command from the command line or from ... googler is a power tool to perform Google Search (Web or News) from
     the ...

 8.  #googler hashtag on Instagram • Photos and Videos
     https://www.instagram.com/explore/tags/googler/
     15.8k Posts - See Instagram photos and videos from 'googler' hashtag.

 9.  googler - Google from the command-line - Ubuntu Manpage
     http://manpages.ubuntu.com/manpages/bionic/man1/googler.1.html
     googler is a command-line tool to search Google (Web & News) from the terminal. Google site search works too. googler shows the title, URL and text
     context for ...

 10. Debian -- Details of source package googler in sid
     https://packages.debian.org/source/unstable/googler
     The following binary packages are built from this source package: googler: Power tool to Google (Web & News) and Google Site Search from the
     terminal ...

googler (? for help) o 6-10
googler (? for help) handling https://www.fosslinux.com/25334/googler-google-search-from-your-linux-terminal.htm
handling https://chocolatey.org/packages/googler
handling https://www.instagram.com/explore/tags/googler/
handling http://manpages.ubuntu.com/manpages/bionic/man1/googler.1.html
handling https://packages.debian.org/source/unstable/googler
```

## Communicating additional info using env vars

Env vars are your friends. You can use a static handler script, writing to, say `$GOOGLER_OUTPUT_FILE`. Then you basically use the same script as the above, except calling `googler` with `GOOGLER_OUTPUT_FILE="$tmpfile"`. You can adapt the demo to this scenario.

---

So basically everywhere I look the problem is solvable. It requires a tiny bit of shell magic for sure, but the "coprocess" part is way more advanced than passing the tmpfile around, and it seems to me that the coprocess part is pretty much necessary for your use case, so the barrier of entry is already high enough.

PR #329, opened 2020-04-02
Strip trailing comma from metadata

zmwangx:
Without this patch we could see awkward output like

```
  3.  iNEWS FTP/FTPS Server Specification - Avid
       http://resources.avid.com/SupportFiles/attach/Broadcast/iNEWS-FTP-FTPS-ServerProtocol.pdf
       Sep 25, 2017,
       FTP (File Transfer Protocol) interaction. The iNEWS FTP Server supports a subset of the FTP commands described in the FTP specification and ...
```

### DIFF
--- a/googler
+++ b/googler
@@ -2234,7 +2234,7 @@ class GoogleParser(object):
                     abstract = abstract + childnode.text.replace('\n', '')
                 try:
                     metadata = div_g.select('.f').text
-                    metadata = metadata.replace('\u200e', '').replace(' - ', ', ').strip()
+                    metadata = metadata.replace('\u200e', '').replace(' - ', ', ').strip().rstrip(',')
                 except AttributeError:
                     metadata = None
             except (AttributeError, ValueError):


googler
```diff
@@ -2235,6 +2235,8 @@ class GoogleParser(object):
                 try:
                     metadata = div_g.select('.f').text
                     metadata = metadata.replace('\u200e', '').replace(' - ', ', ').strip()
+                    if metadata.endswith(','):
+                        metadata = metadata.rstrip(',')
```
jarun, line 2239: What happens if we just `metadata = metadata.rstrip(',')` without the check? If there's no `,` at the end there would be no change to metadata, right?

zmwangx: Yeah, this is what happens when I write down the first line without even thinking about what should come next.

zmwangx: Fixed.

jarun: It's OK. :+1:

jarun:
Thank you!

PR #330, opened 2020-04-02
Update asciinema shot to screencast in animated SVG

zmwangx:
Just want to share something I spent some time working on. So recently I discovered this cool thing -- terminal screencasts as animated SVG. The whole > one minute of it is merely ~80.6KiB gzipped compared to >400KiB of our current PNG screenshot.

Unfortunately this can't be merged because performance is absolute garbage on Chrome for whatever reason (works very well on Firefox and Safari). Whole page is made janky as hell and one frame is rendered every five seconds or something.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,11 @@
 </p>
 
 <p align="center">
-<a href="https://asciinema.org/a/85019"><img src="https://i.imgur.com/EbZof9q.png" alt="Asciicast" width="734"/></a>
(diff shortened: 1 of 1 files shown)


zmwangx:
Isn't this beautiful -- when it works.

jarun:
I didn't have `imgp` when we uploaded the PNG I guess. ;) This really looks cool!

zmwangx:
The current PNG is auto generated by asciinema.org. And there’s not much you can do with a lossless image of a certain size.

Btw, this doesn’t work well for crazy long demos due to performance but if you have short demos it would be perfect. https://github.com/nbedos/termtosvg if you’re interested.

jarun:
Ii think our current demo length is totally justified.

Issue #281, opened 2019-04-21
Always need to scroll back on short terminals


zmwangx:
This is a somewhat frequent complaint (#207, #218, and scattered on hijacked threads). 

There's a fairly obvious solution for this — a goddamn pager! (For some reason I haven't thought of this until now, maybe because of the XY problem — users were always suggesting a dynamic number of results, which is both unnecessarily hard to implement and a poor fit for our paging model, and I was caught up in why that's a bad idea instead of tackling the root cause.)

More concretely, we just need to pipe output into `less -EFX` (`-E` — `--QUIT-AT-EOF`; `-F` — `--quit-if-one-screen`; `-X` — `--no-init`). A tried and true paging technique.

There are two problems with `less -EFX` that are worth mentioning:

1. If user quits early, the rest of the output is lost. We can of course implement a "repeat" command (`r` is conveniently still available) to repeat the current page's results, which largely mitigates the problem. This does mean we can't make paging the default, though. It has to be behind an option (`-P`, `--pager` should be good enough), and discoverability would suffer.

   Discoverability is a really hairy problem for a program that pledges to leave no trace on user's disk. I guess we could suggest the option *once* in each interactive session where the output is ~longer~ way longer than a screenful. That shouldn't be too annoying for people who don't want it, right?

2. `less` is universal enough, with one catch: BusyBox's `less` doesn't have `-F` or `-X` [1] (I'm not even sure if it supports ANSI escapes, haven't tried). A simple preflight with `--help` should shake that out though. I'm not aware of any other `less` implementation except the most common one found in both Linux and BSD.

@jarun Let me know your thoughts.

[1] https://busybox.net/downloads/BusyBox.html#less

zmwangx:
Unrelated: I haven't forgot about the installer script I talked about the other day. I already wrote it a few days ago, but haven't found time to satisfactorily test it yet.

jarun:
I use `less` to show output in some screens in `nnn` but this is a completely different case. I don't know anyone who doesn't set the minimum size of his/her terminal carefully. That's kind of fixed for users.

So where's the problem in using a namesake alias to limit the number of results shown per page? E.g.:

    alias googler=googler -n 6
You can even use auto-completion.

[...]

PR #339, opened 2020-05-08
Implement -r, --reverse option

zmwangx:
Related #281.

```
$ /googler -r google

 5.  Google - YouTube
     https://www.youtube.com/Google
     Experience the world of Google on our official YouTube channel. Watch videos about our products, technology, company happenings and more. Subscribe
     to ...

 4.  Google - About Google, Our Culture & Company News
     https://about.google/
     Stay up to date with Google company news and products. Discover stories about our culture, philosophy, and how Google technology is impacting others.

 3.  Browse All of Google's Products & Services - Google
     https://about.google/products/
     Browse a list of Google products designed to help you work and play, stay organized, get answers, keep in touch, grow your business, and more.

 2.  The Keyword | Google
     https://www.blog.google/
     Discover all the latest about our products, technology, and Google culture on our official blog.

 1.  Google
     https://www.google.com/
     Search the world's information, including webpages, images, videos and more. Google has many special features to help you find exactly what you're
     looking ...
```

### DIFF
--- a/googler
+++ b/googler
@@ -2683,7 +2683,7 @@ class GooglerCmd(object):
                 self.warn_no_results()
             else:
                 sys.stderr.write(prelude)
-                for r in self.results:
+                for r in reversed(self.results) if self._opts.reverse else self.results:
                     r.print()
 
     @require_keywords
@@ -3397,6 +3397,8 @@ def parse_args(args=None, namespace=None):
            help='custom script or cli utility to open results')
     addarg('--show-browser-logs', action='store_true',
            help='do not suppress browser output (stdout and stderr)')
+    addarg('-r', '--reverse', action='store_true',
+           help='reverse printing order of results')
     addarg('--np', '--noprompt', dest='noninteractive', action='store_true',
            help='search and exit, do not prompt')
     addarg('-4', '--ipv4', action='store_const', dest='address_family',


jarun:
Not that way. I meant `r` as a prompt sub-command. So at the prompt if you press `r` the results get reversed. However, when we press `n` or `p` results should show in the order we fetch.

zmwangx:
Kinda confused here. Reversing the order is for people with short terminals and want to see the more relevant results without scrolling. With that in mind, why would they use a separate command to only temporarily reverse the order? Seems about as much trouble as scrolling back. I'd expect people who like this behavior to permanently have it enabled.

jarun:
> I'd expect people who like this behavior to permanently have it enabled.

When you go to the second/third page, all results are equally relevant. The problem is having to scroll up the top results.

zmwangx:
Sure, order doesn’t actually matter on pages other than the first, which means reverse is just as good as default. But having to run a separate command to reverse the order of the first page every time is still about as much trouble as scrolling back. “I'd expect people who like this behavior to permanently have it enabled.” still stands.

jarun:
It helps this way:

We don't work like a pager. We print all results and that may exceed a screenful. With prompt option `r`, _in a page_, one can read the last visible results first and then press `r` to see the topmost results.

Many people are not aware how to scrollback without the mouse.

Ordering the results in reverse doesn't solve the problem.

zmwangx:
Hmm, ten results won’t necessarily fit onto two pages either, and even if they do, if you need to open anything hidden from view you need to scroll back regardless, so I don’t think “avoid scrolling back at all costs” is what users want, nor is it what they’ll get with an r command.

Meanwhile, at least one user expressed wish for a permanent reversal.

Anyway, I think we’re disagreeing on what’s a less bad compromise on a problem neither of us are particularly concerned about ourselves? I guess my preferred solution then is not choose a compromise (unless you want to use the command yourself.)

jarun:
Yes, I do think we can give it a rest. :+1: 
I use a large enough terminal and reasonable enough `-n`. Given people who use `googler` plan to sue it for the long term, they should adjust these things like they set up their terminal or editor.

PR #341, opened 2020-05-24
Fix extraction of matched keywords

zmwangx:
At some point Google apparently switched to the em tag (at least in the version
I'm served).

### DIFF
--- a/googler
+++ b/googler
@@ -2269,7 +2269,7 @@ class GoogleParser(object):
                     if 'f' in childnode.classes:
                         # .f is handled as metadata instead.
                         continue
-                    if childnode.tag == 'b' and childnode.text != '...':
+                    if childnode.tag in ['b', 'em'] and childnode.text != '...':
                         matched_keywords.append({'phrase': childnode.text, 'offset': len(abstract)})
                     abstract = abstract + childnode.text.replace('\n', '')
                 try:


zmwangx:
Btw, I caught this while thinking about rewriting the test suite to be not utterly useless (I surveyed issues within the last two years, and all regression PRs, apparently the CI test never caught anything...)

Will have a PR maybe tomorrow, or tonight if I'm motivated enough.

jarun:
Thank you!

BTW, I checked the Auto-Lock issue and am using the new way to do it in `nnn` now. If ti works as expected, I will update `googler as well.

Details here: https://github.com/dessant/lock-threads-app/issues/2

zmwangx:
Interesting, so googler is only grandfathered in at this point...

jarun:
> grandfathered in

I didn't get it. If you mean de-prioritized, that's not the case. The auto-lock wasn't working for `nnn` so I raised the issue. I want to confirm on `nnn` that it works and then I will apply the changes in `googler`.

zmwangx:
I was saying the lock app works for googler only because it was installed in the past, new installations don't work anymore. ("Grandfathered in" example: I'm grandfathered into <insert telecom>'s $<insert dollar amount> plan, which doesn't exist anymore if you sign up today.)

jarun:
Ahh OK. You are reading literature too much. ;)
I am kind of sucked into office work nowadays. Not even thinking other things in detail.

PR #343, opened 2020-05-24
Testing overhaul

zmwangx:
Rewrite test suite based on survey of past issues and regressions.

Use pytest instead of homegrown "framework". Use structured output to simplify
programmatic testing.

Migrate to GitHub Actions.


zmwangx:
Note: to properly investigate failures we sometimes might need to get our hands on the raw HTML response instead of truncated strings from pytest, so ideally we need to package up the raw HTML responses as artifact. This is already possible right now with --debug (to make things easier we can set a special TMPDIR, then zip up the whole thing at the end). However, I'm saving this work for later if the need actually arises.

zmwangx:
And... the PR doesn't trigger GitHub Actions (guess it has to be in the branch first). Here's how the workflow turned out on my fork: https://github.com/zmwangx/googler/actions/runs/114040792

There should be zero setup other than merging.

jarun:
CI is failing. Please take a look.

zmwangx:
CircleCI is failing because I deleted CircleCI.

jarun:
Cool! Is it possible to have a button for status in Readme.md?

zmwangx:
Added badge.

jarun:
Thank you!

zmwangx:
Cool, passed on master. One more step: need to disable CircleCI, or it will keep reporting build status to GitHub, result in commits marked as failure, and send email 😃

jarun:
Will do. Also how about auto-package generation on new release? Would that work or we need to add something?

zmwangx:
Oh I forgot about that. I'll add it back after taking a shower.

jarun:
Sure thing!

PR #344, opened 2020-05-26
Restore daily tests and packaging on tags

zmwangx:
Packaging is slow (takes ~20min) and I had to iterate a few times due to various stupid mistakes, so I eventually let it run overnight. Here's a successful build https://github.com/zmwangx/googler/runs/707042399?check_suite_focus=true, and here's the automated release created from that: https://github.com/zmwangx/googler/releases/tag/v4.1.packaging_test (The workflow only creates a draft release, so you can still edit it before publishing.)

### DIFF
--- a/.github/workflows/build-and-upload-artifacts.yml
+++ b/.github/workflows/build-and-upload-artifacts.yml
@@ -1,0 +1,32 @@
+name: build and upload artifacts
+on:
+  push:
+    tags:
+      - "v*"
+jobs:
+  build-and-upload-artifacts:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+      - name: Install PackageCore
+        run: |
+          python -m pip install --upgrade pip
+          pip install packagecore
+      - name: Build artifacts with PackageCore
+        run: |
+          version=${GITHUB_REF##*/v}
+          packagecore -o dist/ "$version"
+      - name: Create draft release with artifacts
+        run: |
+          tag=${GITHUB_REF##*/}
+          asset_options=()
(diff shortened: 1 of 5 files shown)


zmwangx:
Btw, the list of distros to package for might need to be reviewed. For instance, debian8 and ubuntu14.04 are irrelevant or soon to be irrelevant at this point; they don't even ship a Python we support (py34 EOL'ed last year).

I didn't do the review because I'm not exactly an enthusiast and just go with Ubuntu/Debian myself.

zmwangx:
Just tagged on another largely unrelated commit here. When reviewing the examples I just realized we actually have a `-w, --site` option for the `site:` syntax.

jarun:
I am aware of the debian8 and the ubuntu 14.0 things. But they are not EOL yet and there are ways to get newer versions of python on them (personal PPAs).

Let's just wait for a while.

jarun:
Thank you!

zmwangx:
Ubuntu 14.04 EOL'ed last April though.

jarun:
14.04.6 is still supported: https://wiki.ubuntu.com/Releases

zmwangx:
Oh, I thought the extended maintenance was for paid enterprise customers only.

jarun:
Even in that case, I do not know what our policy should be. Ubuntu is not paying us for sure! ;)

zmwangx:
I mean if it works, great; if it doesn’t, I’m certainly not gonna fix it for specifically for Ubuntu 14.04. Like RedHat, Canonical make money selling extended support, so the burden of supporting their outdated system falls on them. (I don’t think any enterprise uses googler in any capacity, so just making a general remark.)

Btw, the extended support is only free for personal use on up to three machines, and I seriously doubt any person’s subjecting them to outdatedness on this level on a non-server.

jarun:
OK. You have a point. We can remove these.

PR #345, opened 2020-05-26
Prefer IPv4 by default

zmwangx:
I thought about this when I implemented -4 and -6 but somehow convinced myself that picking a network interface out of order is a slight overreach.

I reconsidered just now, and now that I'm more convinced that IPv6 is a problem, I think this is a positive change. Working out of the box is better than printing a long error message, embedding instructions in it, and assuming users actually read it (some simple won't) still requiring an extra option or alias.


jarun:
Thank you!

PR #346, opened 2020-05-26
README: pull usage examples to the top

### LINKED ISSUE #336: Pulling examples to the top of README?
Today’s issue #335 got me thinking, the README is huge right now, and whenever I see an overwhelming README that’s almost like Bash Reference Manual on one single HTML page, I want to see examples. The demo screencast is a start but not everyone is into watching “videos”.

What about pulling the examples section up? I’m thinking between the intro and the TOC (so it becomes a TOC for detailed, technical documentation).

Could also use an example showcasing search phrases being split into arbitrary command line arguments.

zmwangx:
The README is too damn long at this point, so prioritize the human friendly part.

Also added a few more examples showcasing --from/--to, --videos, and positional argument chaining.

Closes #336.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,98 @@
 
 `googler` isn't affiliated to Google in any way.
 
+Here are some usage examples:
+
+1. Google **hello world**:
+
+        $ googler hello world
+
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

PR #351, opened 2020-07-18
README: add note on GUI browser integration in WSL

zmwangx:
I was using googler in WSL myself and had to Google this one, so might as well document it.


jarun:
Thank you!

PR #353, opened 2020-07-18
Make debug mode even more informative

zmwangx:
- Check new version (cached for 24 hours);
- Print `platform.platform()`.


jarun:
Thank you!

PR #354, opened 2020-07-18
Introduce hidden debug option --parse to parse dumped HTML

zmwangx:
There used to be a separate directory called `devbin` or something with a script to load dumped HTML and call the parser, and the parser only.

This option makes it easy to directly load dumped HTML with googler and trace throughout the entire parsing and rendering cycle.

I introduced this to help me fix two other bugs (PRs to follow).


jarun:
Thank you!

PR #355, opened 2020-07-18
Fix sitelinks parsing

zmwangx:
Sitelink abstract now bears the following structure

```html
<div class="s">
  <div class="st" style="overflow:hidden;width:220px">Sign in. Use your Google Account ... Use Guest mode to sign in ...<br></div>
</div>
```

at least for me. I think the parser has been broken for quite a while in this regard.


jarun:
Thank you!

PR #357, opened 2020-07-18
Do not print an additional blank line when abstract is empty

zmwangx:
Not actually sure why we never changed this... But I happened to be developing against a request that resulted in an empty abstract today, and the double blank line sure looked pointless (especially when it's repeated due to the result duplication bug).

Before (yeah, this is without the duplicate suppression):

```console
$ ./googler python3.5 eol

 1.  Is there official guide for Python 3.x release lifecycle? - Stack ...
     https://stackoverflow.com/questions/40655195/is-there-official-guide-for-python-3-x-release-lifecycle


 2.  Is there official guide for Python 3.x release lifecycle? - Stack ...
     https://stackoverflow.com/questions/40655195/is-there-official-guide-for-python-3-x-release-lifecycle


 3.  17. Development Cycle — Python Developer's Guide
     https://devguide.python.org/devcycle/
     A branch less than 5 years old but no longer in maintenance mode is a ... For reference, here are the Python versions that most recently reached their
     end-of-life: ...
```

After:

```console
$ ./googler python3.5 eol

 1.  Is there official guide for Python 3.x release lifecycle? - Stack ...
     https://stackoverflow.com/questions/40655195/is-there-official-guide-for-python-3-x-release-lifecycle

 2.  Is there official guide for Python 3.x release lifecycle? - Stack ...
     https://stackoverflow.com/questions/40655195/is-there-official-guide-for-python-3-x-release-lifecycle

 3.  17. Development Cycle — Python Developer's Guide
     https://devguide.python.org/devcycle/
     A branch less than 5 years old but no longer in maintenance mode is a ... For reference, here are the Python versions that most recently reached their
     end-of-life: ...
```

(As an aside, it would be nice if we could actually display the text in a featured snippet, but those snippets are a little bit too liberal in format to reliably parse, so I guess we're stuck with a bare title and URL for now).

### DIFF
--- a/googler
+++ b/googler
@@ -2483,22 +2483,24 @@ class Result(object):
             else:
                 print(' ' * (indent + 5) + metadata)
 
-        fillwidth = (columns - (indent + 6)) if columns > indent + 6 else len(abstract)
-        wrapped_abstract = TrackedTextwrap(abstract, fillwidth)
-        if colors:
-            # Highlight matches.
-            for match in matches or []:
-                offset = match['offset']
-                span = len(match['phrase'])
-                wrapped_abstract.insert_zero_width_sequence('\x1b[1m', offset)
-                wrapped_abstract.insert_zero_width_sequence('\x1b[0m', offset + span)
+        if abstract:
+            fillwidth = (columns - (indent + 6)) if columns > indent + 6 else len(abstract)
+            wrapped_abstract = TrackedTextwrap(abstract, fillwidth)
+            if colors:
+                # Highlight matches.
+                for match in matches or []:
+                    offset = match['offset']
+                    span = len(match['phrase'])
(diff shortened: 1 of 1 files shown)


zmwangx:
Btw we've racked up a sizable list of fixes and improvements so it might be time to consider a release.

jarun:
Thank you!

That's true. I'll make a release soon.

jarun:
I will make the release sometime next week.

jarun:
packaging for arch linux is still failing. See https://github.com/BytePackager/packagecore/issues/204

zmwangx:
This has caused way too much trouble for it worth, my impression is it's been failing half of the time when we used it. I'm also like 99.9% sure no arch user install the package this way -- there's just no reason to not use AUR which should always be on the latest version. Why don't we just remove it.

jarun:
Agreed. Will do this next week.

PR #362, opened 2020-09-04
fix missing closing quote

0xF4D3C0D3:
it's really trivial typo though, someone needs to PR it I think.
if it's too trivial to be merged, ignore it please

thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Here are some usage examples:
 
 18. Quote multiple search keywords to auto-complete (using completion script):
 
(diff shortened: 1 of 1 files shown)


zmwangx:
No, this is intended. `<TAB>` indicate where you press tab to get autocomplete results, basically serving as a cursor here. The closing quote hasn’t been tapped at that point. But thanks for the interest anyway.

0xF4D3C0D3:
oh I got it, I missed that point thx!

PR #371, opened 2020-09-23
fixed stdin for piped commands

PatxiofromAlphensign:
Added `stdin.read()` check if no `opts.keywords` are not given. To be well used for `echo googler | googler`

### DIFF
--- a/googler
+++ b/googler
@@ -3610,6 +3610,13 @@ def main():
         Result.urlexpand = True if os.getenv('DISABLE_URL_EXPANSION') is None else False
         GooglerCmd.colors = colors
 
+ 
+        frmStdin = None
+        if len(opts.keywords) < 1:
+            frmStdin =  [sys.stdin.read()]
+            opts.keywords += frmStdin
+
+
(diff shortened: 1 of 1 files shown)


googler
```diff
@@ -3610,6 +3610,13 @@ def main():
         Result.urlexpand = True if os.getenv('DISABLE_URL_EXPANSION') is None else False
         GooglerCmd.colors = colors
 
+ 
+        frmStdin = None
+        if len(opts.keywords) < 1:
+            frmStdin =  [sys.stdin.read()]
```
jarun, line 3616: `read()` is a blocking call. With this `googler` hangs when there is no input:

```
1 ~/GitHub/googler$ ./googler      
^C
Interrupted.
1 ~/GitHub/googler$
```

jarun (changes requested):

zmwangx:
Late reply, but if you really need this you'll have to check stdin isn't a tty first (`sys.stdin.isatty()`). But I'd recommend command substitution, probably assigning to a variable first.

Issue #372, opened 2020-10-01
Stopped working?


097115:
- Output of `googler -d`:

 ```
[DEBUG] googler version 4.2
[DEBUG] Python version 3.7.4
[DEBUG] Platform: Darwin-15.6.0-x86_64-i386-64bit
[DEBUG] Connecting to new host www.google.com
[DEBUG] Opened socket to 216.58.215.68:443
[DEBUG] Fetching URL /search?ie=UTF-8&oe=UTF-8&q=test&sei=YhfWbCYXQL2KPfSS1sfGbA
[DEBUG] Cookie: CGIC=Igl0ZXh0L2h0bWw
[DEBUG] Response body written to '/var/folders/fm/flpqjgss1mz5hhs92sj_6fm40000gn/T/googler-response-h3lbic2q.html'.
No results.
If you believe this is a bug, please review https://git.io/googler-no-results before submitting a bug report.
```

- Link to the response body:

https://gist.github.com/097115/821c10653b18cc83d333329ab8371e9c

- Details of operating system, Python version used:

Both Ubuntu and macOS, both python 3.5.2 and 3.7.4. Worked just a few hours ago, and now suddenly returns no results (both 4.2 and the latest master).

**Comment by hooksoft** [comment]
Same, first time using this tool a few minutes ago.

**Comment by matthewstrasiotto** [comment]
Weirdly enough, for me- the following works:

```bash
googler --debug aaaa
```
```
[DEBUG] googler version 4.2
[DEBUG] Python version 3.8.5
[DEBUG] Platform: Linux-5.4.0-48-generic-x86_64-with-glibc2.2.5
[DEBUG] Connecting to new host www.google.com
[DEBUG] Opened socket to 172.217.167.100:443
[DEBUG] Fetching URL /search?ie=UTF-8&oe=UTF-8&q=aaaa&sei=Dz1GrARhEeu1_XCFwrImsQ
[DEBUG] Cookie: 1P_JAR=2020-10-02-03
[DEBUG] Response body written to '/tmp/googler-response-oqc8_miv.html'.

 1.  Contact Us
     https://www.aaaa.com.au/contact-us/
     Contact Us. National Office 7-8 Bastow Place Mulgrave, Victoria ...

     1a. Contact Us
         https://www.aaaa.com.au/contact-us/
         Contact Us. National Office 7-8 Bastow Place Mulgrave, Victoria ...

     1b. Join AAAA
         https://www.aaaa.com.au/join-aaaa/
[...]

PR #377, opened 2020-10-05
Introduce a protocol for Google-breaks-everything incidents

zmwangx:
There have been enough Google-breaks-everything shitstorms, it's time to introduce a better protocol to reduce user confusion during these trying periods.

Here's the protocol:

- Once a report is confirmed and proven to be non-transient (e.g. still reproducible after 24 hours), regardlessly of whether a fix can be made available immediately, the epoch in `googler` and `info.json` should be advanced to the date of the report (in UTC), and the issue number recorded. `info.json` should look like:

  ```json
  {
      "epoch": "20201001",
      "tracking_issue": 372,
      "fixed_on_master": false,
      "fixed_in_release": false
  }
  ```

  At this point users would see a message like this:

  ```console
  $ googler hello  # Issue acknowledged, not fixed.
  No results.
  Your version of googler is broken due to Google-side changes.
  The issue is tracked at https://github.com/jarun/googler/issues/372.
  ```

- Once a fix is committed to master, `fixed_on_master` should be set to `true`:

  ```json
  {
      "epoch": "20201001",
      "tracking_issue": 372,
      "fixed_on_master": true,
      "fixed_in_release": false
  }
  ```

  At this point users would see a message like this:

  ```console
  $ googler hello  # Fixed on master, not released.
  No results.
  Your version of googler is broken due to Google-side changes.
  The fix has been pushed to master, pending a release.
  Please download the master version https://git.io/googler or wait for a release.
  ```

- Once the fix is released, `fixed_in_released` should be updated to the version tag:

  ```json
  {
      "epoch": "20201001",
      "tracking_issue": 372,
      "fixed_on_master": true,
      "fixed_in_release": "v4.3"
  }
  ```

  At this point users would see a message like this:

  ```console
  $ googler hello  # Fix released.
  No results.
  Your version of googler is broken due to Google-side changes.
  A new version, v4.3, has been released to address the changes.
  Please upgrade to the latest version.
  ```

I'll create a wiki page for this protocol if we can agree on it.

### DIFF
--- a/googler
+++ b/googler
@@ -90,6 +90,7 @@ except ValueError:
 # Constants
 
 _VERSION_ = '4.2'
+_EPOCH_ = '20201001'
 
 COLORMAP = {k: '\x1b[%sm' % v for k, v in {
     'a': '30', 'b': '31', 'c': '32', 'd': '33',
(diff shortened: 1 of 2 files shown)


jarun:
I'm good with this. We should add the link well-highlighted in the new defect template as well.

jarun:
Thank you!

zmwangx:
> We should add the link well-highlighted in the new defect template as well.

Sorry, which link?

jarun:
> I'll create a wiki page for this protocol if we can agree on it.

To this Wiki page.

zmwangx:
Added https://github.com/jarun/googler/wiki/The-Google-Breaks-Everything-Protocol (sorry for the delay).

The page is for developers (i.e. us) though. There already a user-oriented page (https://github.com/jarun/googler/wiki/What-to-do-when-you-encounter-%22No-results%22) which is linked in the warning message, probably to limited effect.

I do think we should raise the barrier of opening a new issue. I believe GitHub has introduced tools to achieve that to some extent. I’ll look into it later.

PR #380, opened 2020-10-10
Replace em dash in metadata

zmwangx:
Google is now using U+2014 EM DASH instead of U+002D HYPHEN MINUS as separator, at least sometimes.


jarun:
Thank you!

PR #381, opened 2020-10-10
Fix metadata parsing

zmwangx:
Metadata parsing is so damn unreliable now, that the fields detection already
broke after a few days. I've no choice but to use a magic class now, which
could break any second.

Also, turns out this flaky metadata node detector also has the tendency to pick
up non-metadata crap like sublinks and in certain cases, entire accordians of
stuff (like announcements under certain covid results). Slap some bandages on
top to exclude those.

### DIFF
--- a/googler
+++ b/googler
@@ -2378,26 +2378,28 @@ class GoogleParser(object):
                         continue
                     for child in node.children:
                         abstract_nodes.append(child)
+                metadata = None
                 try:
                     # Sometimes there are multiple metadata fields
                     # associated with a single entry, e.g. "Released",
                     # "Producer(s)", "Genre", etc. for a song (sample
                     # query: "never gonna give you up"). These need to
                     # be delimited when displayed.
-                    metadata_fields = metadata_node.select_all('div > div[data-ved]')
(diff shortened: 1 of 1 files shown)


jarun:
There's some conflict. Also, I am removing references to Python 3.5 in docs and CI. It's EOL now: https://devguide.python.org/#status-of-python-branches

jarun:
And will add 3.9 to CI.

zmwangx:
Yeah, it's time.

Fixed conflict. This should be the last thing that has to be in a release.

jarun:
:+1:

jarun:
I don't know why. But somehow the latest version seem a little slow. I have to wait for a while before the results show.

Maybe for the next release we can target performance.

zmwangx:
Strange. You might to use cProfile to see what part is the bottleneck. I haven’t noticed anything.

jarun:
I will try in the evening again. May be network issue also.

jarun:
But in general, it won't be a bad exercise to profile and analyze where we stand now.

Issue #382, opened 2020-10-10
Performance


zmwangx:
@jarun mentioned feeling slow, so I did some profiling.

Since the networking part differs by environment and is mostly outside our control (unless we want to get into the HTTP client implementation business), I simply prefetched responses using `--debug`, then profiled `googler --parse <file>` (recall that `--parse` is the hidden developer tool to reproducibly parse a response HTML file).

Here's an overview of some queries I performed and the most important time measurements (`parse` is the total amount of time spent parsing, `parse_html` is the DOM-building part of that — using my DOM building library [dim.py](https://github.com/zmwangx/dim)):

| query            | size     | #nodes | #elements | parse  | parse_html |
| ---------------- | -------- | ------ | --------- | ------ | ---------- |
| apple            | 694461B  | 2058   | 1617      | 0.122s | 0.062s     |
| covid            | 1459581B | 6376   | 4690      | 0.344s | 0.190s     |
| forbidden palace | 582147B  | 1955   | 1542      | 0.118s | 0.060s     |
| googler          | 323438B  | 1240   | 956       | 0.079s | 0.038s     |
| linux            | 490538B  | 1839   | 1439      | 0.108s | 0.056s     |
| microsoft        | 458318B  | 1709   | 1338      | 0.106s | 0.052s     |
| white house      | 632724B  | 2501   | 1967      | 0.153s | 0.076s     |

("size" is the size of the decompressed HTML response; "#nodes" is the number of HTML nodes in the response, and "#elements" is the number of element nodes among those.)

Times are as measured by `cProfile`'s default config (wall clock) on a middle-of-the-road Core i7-8700B @ 3.20GHZ CPU.

As we can see, the parser takes ~100ms (give or take 50%) total on a typical query on my so-so CPU, with about 50% of the time spent building the DOM then another 50% extracting info, which I deem pretty reasonable. It must be orders of magnitude slower than, say Blink/Webkit/Gecko/html5ever (oh and our parser is not really an HTML5 parser and craps its pants on tag soup), but this is a pure Python parser optimized for modularity and readability (hopefully), so we can't expect too much from it. However there are pathological cases like "covid" which takes ~350ms, here we're inching into the embarrassingly slow territory. (Do note "covid" is a truly pathological case — there's so much more stuff returned, including some unique stuff, as evidenced by the sheer number of nodes/elements, which is at least 2x other queries.)

[...]

PR #383, opened 2020-10-10
Add time_it decorator to profile the most time-consuming methods in debug mode

zmwangx:
Could be useful for getting a sense of the rough performance of things without pulling out a full blown profiler.

Timing example on a really crappy Linux box:

Screen Shot 2020-10-10 at 10 06 00 PM

Related: #382.


jarun:
Thank you!

PR #386, opened 2020-10-18
Do not sabotage stdin of url_handler

### LINKED ISSUE #385: Interactive script as url-handler?
Would be great if one could use an interactive script as `--url-handler` option.

Example workflow:

  - run `googler`, get a list of results
  - select a result, it gets passed to a script
  - script inspects the result (via `curl`, for instance) and asks a user, if they want to proceed
  - script acts depending on the received reply.

Or another take:

  - run `googler`, get a list of results
  - select a result, it gets passed to a script
  - script asks what browser (or other program) you want to open the result in
  - script acts depending on the received reply.

For now, as far as I can tell, `googler` doesn't allow to achieve that (but then again, may be I'm just plain wrong?)

zmwangx:
Makes writing interactive handlers easier.

Fixes #385.

### DIFF
--- a/googler
+++ b/googler
@@ -35,6 +35,7 @@ import shutil
 import signal
 import socket
 import ssl
+import subprocess
 from subprocess import Popen, PIPE, DEVNULL
 import sys
 import textwrap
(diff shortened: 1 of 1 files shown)


jarun:
Thank you!

Issue #392, opened 2021-01-15
googler [V. 4.3.1] : no results  as on  Jan_15_2021


har5ha:
Any help would be appreciated. Thanks! in advance.

```
$googler --debug bash exit codes
[DEBUG] googler version 4.3.1
[DEBUG] Python version 3.8.6
[DEBUG] Platform: Linux-5.7.0-kali1-amd64-x86_64-with-glibc2.29
[DEBUG] Connecting to new host www.google.com
[DEBUG] Opened socket to 172.217.167.164:443
[DEBUG] new_connection completed in 0.133s
[DEBUG] Fetching URL /search?ie=UTF-8&oe=UTF-8&q=bash+exit+codes&sei=DJHJdZAFTLmn5LR6gWMJwg
[DEBUG] Cookie: 1P_JAR=2021-01-15-06
[DEBUG] fetch_page completed in 1.379s
[DEBUG] Response body written to '/tmp/googler-response-meu360rs.html'.
[DEBUG] parse completed in 0.036s
No results.
[DEBUG] Fetching https://raw.githubusercontent.com/jarun/googler/master/info.json for project status...
If you believe this is a bug, please review https://git.io/googler-no-results before submitting a bug report.
googler (? for help)
```
Gist : https://gist.github.com/har5ha/47763f2f39ed3fb85f2c0a128331c83c

Attachment  HTML :
[googler-response-meu360rs.html.txt](https://github.com/jarun/googler/files/5818971/googler-response-meu360rs.html.txt)

raz-bn:
Got the same issue when trying to use v4.3 and v4.2

morganmay:
```
[DEBUG] googler version 4.3.1
[DEBUG] Python version 3.9.1
[DEBUG] Platform: Linux-5.10.7-arch1-1-x86_64-with-glibc2.32
[DEBUG] Connecting to new host www.google.com
[DEBUG] Opened socket to 108.177.111.106:443
[DEBUG] Fetching URL /search?ie=UTF-8&oe=UTF-8&q=test&sei=DIy1U_GLQAW8rUJtW5FF5w
[DEBUG] Cookie: 1P_JAR=2021-01-16-21
[DEBUG] Response body written to '/tmp/googler-response-dyyo_3gu.html'.
No results.
[DEBUG] Fetching https://raw.githubusercontent.com/jarun/googler/master/info.json for project status...
If you believe this is a bug, please review https://git.io/googler-no-results before submitting a bug report.
```
[...]

PR #393, opened 2021-01-20
Fix html class of result nodes

amlatyrngom:
This PR is a quick fix for issue #392. I've manually tested on a few inputs to make sure that it works. However, I can't guarantee that the fix generalizes to all inputs.


jarun:
Thank you!

Issue #402, opened 2021-05-27
Googler crashes when the --tld value is not set explicitly


Battery233:
## My environment 💻
googler: v 4.3.2 (latest)
py: v.3.9.5
MacOS 11.4 & Zsh
locale: All UTF8 for googler
Also, proxy is enabled in my environment:
export http_proxy=http://127.0.0.1:12639
export https_proxy=http://127.0.0.1:12639

### First, I execute `➜ ~ googler -d --tld=hk test`:
It works fine ☑️
log📃⬇️
[DEBUG] googler version 4.3.2
[DEBUG] Python version 3.9.5
[DEBUG] Platform: macOS-11.4-x86_64-i386-64bit
[DEBUG] Connecting to proxy server 127.0.0.1:12639
[DEBUG] Tunnelling to host www.google.com.hk
[DEBUG] Opened socket to 127.0.0.1:12639
[DEBUG] new_connection completed in 0.106s
[DEBUG] Fetching URL /search?ie=UTF-8&oe=UTF-8&q=test&sei=hMypT5SJTLC9hXt1fgve4w
[DEBUG] Cookie: 1P_JAR=2021-05-27-04
[DEBUG] fetch_page completed in 1.856s
[DEBUG] Response body written to '/var/folders/v4/6p2nxvvn2g5f6gzb__pfq3dh0000gn/T/googler-response-ol0upico.html'.
[DEBUG] parse completed in 0.032s

### Then I execute `➜  googler -d test`:
It crashed 💥
It seems that two redirecting are involved in this case and the second one failed.

Since it is redirecting from google.com.hk to google.com.hk, the statement at
https://github.com/jarun/googler/blob/71c89fdee6ac96f18b354c2ec255a6e68d5cdfc1/googler#L2269
is not executed.

This crash is reproducible only when the local proxy is enabled.

log📃⬇️
[DEBUG] googler version 4.3.2
[DEBUG] Python version 3.9.5
[DEBUG] Platform: macOS-11.4-x86_64-i386-64bit
[DEBUG] Connecting to proxy server 127.0.0.1:12639
[DEBUG] Tunnelling to host www.google.com
[DEBUG] Opened socket to 127.0.0.1:12639
[DEBUG] new_connection completed in 0.059s
[DEBUG] Fetching URL /search?ie=UTF-8&oe=UTF-8&q=test&sei=SolO8XrVTM6iz2j83yZPPA
[DEBUG] Cookie: 1P_JAR=2021-05-27-03
[...]

PR #403, opened 2021-05-27
Fixed a bug which will cause crash when local proxy is enabled

Battery233:
A quick fix for issue #402 and it worked.


jarun:
Thank you!

PR #421, opened 2021-10-29
Added a printurl option that prints the search url

rgardner4012:


jarun:
You can always use `-d` and grep the URL. What do you need a new flag for?

PR #424, opened 2021-11-04
include alias inside shell completion

CantoroMC:
I just decided to use googler, after being a fan of nnn.
I have seen the shell alias provided in `goggler_at` but I think that instead of
remembering and sourcing aliases, having shell completion providing those sites 
would be more handy.
I use zsh and I have only a limited knowledge on his completion system.
I have reported the site provided for `-w` and `-N` in `googler_at` 
in `auto-completion/zsh/_googler`, so now pressing Tab after the two flags provides
the sites.
Hope this is useful, I will take a look and use googler from here on and maybe I
can further improve this.


jarun:
Shell completions are for program options. Please have this locally. Some packagers install the completion scripts as well and all users may not prefer to have aliases.

PR #426, opened 2021-11-12
Update PROTOCOL_TLS to PROTOCOL_TLS_CLIENT

paxri01:
Added check for 'PROTOCOL_TLS_CLIENT' (Since 3.6) and required options.

I am no python expert, so feel free to modify as desired, it works on my machine. :smiley:


jarun:
Thank you!

